PR #3, opened 2017-11-29
lsp-line: Add faces for hover contents

MaskRay:
So that the face of the text beside `lsp-line-symbol` (called `contents` in `interface Hover`) is also customizable.

### DIFF
--- a/lsp-line.el
+++ b/lsp-line.el
@@ -50,6 +50,14 @@ It is used to know when the window has changed of width.")
 
 (defvar-local lsp-line--timer nil)
 
+(defface lsp-line-contents nil
+  "Face used to highlight the hover's contents."
+  :group 'lsp-ui)
+
+(defface lsp-line-current-contents nil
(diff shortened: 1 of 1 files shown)


topisani:
this stops you from syntax highlighting that line. Its probably the choice id go with too, but we need the option

sebastiencs:
As @topisani said, this prevents to use syntax highlighting. There should be an option to choose whether to use a face or to use syntax highlighting.

MaskRay:
And if `lsp-line-symbol` can be changed to specify other text properties, not just face, i.e. `invisible`

sebastiencs:
This could be set with a different variable containing the text properties ?

MaskRay:
`lsp-line-symbol-text-properties`?

sebastiencs:
Yes :+1:

sebastiencs:
What other properties could be used ? And why would you want to use the property `invisible` ?

PR #4, opened 2017-11-29
Change lsp-line-enable to minor mode

MaskRay:
Do not enable lsp-line by default `(add-hook 'lsp-mode-hook 'lsp-line-enable)`. Instead, let the user choose whether the minor mode should be enabled.


sebastiencs:
Yes, thank you

PR #5, opened 2017-12-02
lsp-ui: Add imenu

MaskRay:
```elisp
(setq-local imenu-create-index-function #'lsp-ui-imenu-create-index)
```

then `M-x imenu`

### DIFF
--- a/lsp-ui.el
+++ b/lsp-ui.el
@@ -33,7 +33,21 @@
 
 ;;; Code:
 
+(require 'imenu)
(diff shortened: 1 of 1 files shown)


sebastiencs:
Nice ! I may be wrong, but this should be in `lsp-mode` ? @vibhavp

vibhavp:
Yes. I've extended this further so as to categorize symbols as per `SymbolKind` values mentioned in protocol. Feedback welcome at #205

PR #16, opened 2017-12-20
Fix gethash on nil.

scturtle:
Signed-off-by: scturtle <scturtle@gmail.com>


topisani:
looking good to me - had this issue too

PR #22, opened 2017-12-30
[lsp-ui-peek] Propagate lsp--cur-workspace when jumping to xref and enable lsp-mode when necessary

jiegec:
With this pr, after jumping outside of project root (e.g. system function definitions), we can still have lsp-mode enabled and the buffer associated with the original lsp--cur-worksapce.

### DIFF
--- a/lsp-ui-peek.el
+++ b/lsp-ui-peek.el
@@ -392,8 +392,13 @@ X."
                             (goto-char 1)
                             (forward-line line)
                             (forward-char column)
-                            (point-marker))))))
+                            (point-marker)))))
+              (current-workspace lsp--cur-workspace))
           (switch-to-buffer (marker-buffer marker))
+          (unless lsp--cur-workspace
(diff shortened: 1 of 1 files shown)


lsp-ui-peek.el
```diff
@@ -392,8 +392,13 @@ X."
...
                             (forward-line line)
                             (forward-char column)
-                            (point-marker))))))
+                            (point-marker)))))
+              (current-workspace lsp--cur-workspace))
           (switch-to-buffer (marker-buffer marker))
+          (unless lsp--cur-workspace
+            (setq lsp--cur-workspace current-workspace))
+          (unless lsp-mode
+            (lsp-mode t))
```
MaskRay: I think `(lsp-mode 1)` is more idiomatic.

MaskRay:
Merged in b62f5fdb5544702d49a004c58276b5dbe8dcddbf

PR #25, opened 2017-12-30
[lsp-ui-peek] Make lsp-ui-peek-mode-map have the highest priority temporarily and deactivate it when aborting or pressing a key not found in the keymap

jiegec:


MaskRay:
This addresses an issue we ran into: `q` does not close the quick-peek in Evil normal state.

sebastiencs:
Sorry for the delay, @jiegec Is it possible to move the code in the function `lsp-ui-peek-mode` ?

jiegec:
Yes, of course. Done in #28.

Issue #21, opened 2017-12-27
Workaround for child-frame's bad behavior under macOS.


jiegec:
I am using Emacs 26 pretest (i.e. 26.0.90) and `lsp-ui-doc.el` is really amazing.
However, similar to what I have run across when debugging `pyim`'s child-frame support for choosing Chinese characters, the child frame might not show any more after a `make-frame-invisible` call. Tested `make-frame-visible` and `set-frame-parameter` with `visibility`, but to no avail. I think this should be a BUG somewhere in Emacs itself, but I hadn't send a bug report to `bug-gnu-emacs`. My temporary fix for `lsp-ui`:

```elisp
diff --git a/lsp-ui-doc.el b/lsp-ui-doc.el
index 358a051..a8c4fa6 100644
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -224,7 +224,11 @@ BUFFER is the buffer where the request has been made."
   (when (lsp-ui-doc--get-frame)
     (lsp-ui-doc--with-buffer
      (erase-buffer))
-    (make-frame-invisible (lsp-ui-doc--get-frame))))
+    (lsp-ui-doc--render-buffer "Hover on symbols to see document." "")
+    (lsp-ui-doc--move-frame (lsp-ui-doc--get-frame))
+    (set-frame-parameter (lsp-ui-doc--get-frame) 'alpha '(0 . 0))
+    ;; (make-frame-invisible (lsp-ui-doc--get-frame))
+    ))
 
 (defun lsp-ui-doc--buffer-width ()
   "Calcul the max width of the buffer."
@@ -312,6 +316,7 @@ SYMBOL STRING."
     (unless (frame-live-p (lsp-ui-doc--get-frame))
       (lsp-ui-doc--set-frame (lsp-ui-doc--make-frame)))
     (lsp-ui-doc--move-frame (lsp-ui-doc--get-frame))
+    (set-frame-parameter (lsp-ui-doc--get-frame) 'alpha '(100 . 100))
     (unless (frame-visible-p (lsp-ui-doc--get-frame))
       (make-frame-visible (lsp-ui-doc--get-frame)))))
```

The problem of this workaround is that, redisplay is too often called, making screen constantly flashing. To refactor this, we need to maintain a visibility state ourselves.

I am writing a PoC for this BUG and will send it to `bug-gnu-emacs` some time later.

[...]

PR #30, opened 2017-12-31
[lsp-ui-doc] Fix issue #21 of frame disappearing by making frame visible before moving it

jiegec:


jiegec:
Emmm sometimes the bug still occur? really weird.

PR #34, opened 2018-01-04
[lsp-ui-doc] refresh frame on theme change

topisani:
When switching themes the frame background would break for me. I'm not sure this advice should be in here, but at least its nice to have the function to delete the frame so it can be reset

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -375,6 +375,7 @@ SYMBOL STRING."
 
 (defun lsp-ui-doc--make-frame ()
   "Create the child frame and return it."
+  (lsp-ui-doc--delete-frame)
   (let* ((after-make-frame-functions nil)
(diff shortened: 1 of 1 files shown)


sebastiencs:
Thanks

hyiltiz:
Had to bisect through 2k lines of emacs config just to pinpoint an innocent looking line `(use-package lsp-ui)`. It was breaking a lot of themes in `doom-themes`, `humanoid-themes`, `leuven` among others although somehow `spacemacs-themes` and `poet` theme still can recover themselves. What's broke in those themes, when `(use-package lsp-ui)` is enabled, is that the cursor highlight is a horrible looking olivegreen color, and the borders of the frame looks black. This happens whether emacs was used with or without the server/daemon.
Getting rid of the line fixed it. Looking for a better workaround that still keeps `lsp-ui`.

PR #35, opened 2018-01-05
Give ‘lsp-ui-mode’ a docstring that describes what the mode function does

phst:


sebastiencs:
Yes the code needs more documentation. Thank you !

PR #41, opened 2018-01-11
Don’t try to enable ‘lsp-ui-doc’ unless child frames are available

phst:


sebastiencs:
Thanks, is it necessary ? Because we already check that child frames are available [there](https://github.com/emacs-lsp/lsp-ui/blob/master/lsp-ui-doc.el#L420)

phst:
Good point. My take is: If enabling the mode doesn't work, it shouldn't be available. Otherwise users get confused (especially if they only get a simple message, not an error), and more code is required at runtime. So my suggestion would be: Factor out lsp-ui-doc into a separate package (which can still be in this repo) and declare the new package incompatible with Emacs 25.

PR #60, opened 2018-01-21
[lsp-ui-peek] Add some more-standard keybindings

gpittarelli:
Adds some more default keybindings for lsp-ui-peek which more closely match the rest of emacs.


sebastiencs:
Thank you !

PR #62, opened 2018-01-21
[lsp-ui-sideline] Check for capability before invoking textDocument/codeAction

Declspeck:
Sideline does not currently check caps before sending `textDocument/codeAction`. This causes an error message with php-language-server every time the cursor is moved.

I also have a question regarding PHP support - would it be possible to get the `lsp-php` repo below the `emacs-lsp` organization when I publish it? I will give it a test run at work next week and push it to GitHub after that if there are no significant problems.


sebastiencs:
@Declspeck For the repo you need to ask to @vibhavp

Declspeck:
Thanks! I'll get in touch when I'm confident that it works!

PR #64, opened 2018-01-22
Fix that peek key bindings don't work under emacs -nw.

scturtle:
Some original key bindings do not works under emacs -nw. Not knowing why.

Other key bindings also have problems with evil-mode even in gui mode:
1. call `lsp-ui-peek-find-references` in normal mode
2. `m-x` than `esc`, still in normal mode, but all keys do not works, no way to exit this minor mode
3. `i` to insert mode, all keys work again


jiegec:
The differences between RET and <return> should be investigated further because I tested that on GUI, the latter is working instead of the former.

scturtle:
Tested with nightly build emacs on linux and mac. The changed key bindings works in both GUI and command-line mode. Maybe there is version specified issue. Need more feedback.

PR #69, opened 2018-01-22
[lsp-ui-imenu] cleanup overlays when refreshing *lsp-ui-imenu*

Sarcasm:
Without this change, section titles were accumulated
everytime the buffer was updated.


sebastiencs:
Ah yes, thanks !

PR #72, opened 2018-01-23
[lsp-ui-peek] Wrap textDocument/definition result in a sequence if not already.

Declspeck:
Language servers are free to reply to `textDocument/definition` with a single `Location` and PHP language server does. This causes a type error when peeking definitions. This wraps the result into a list if it is not a sequence already.


sebastiencs:
Thanks !

PR #83, opened 2018-02-17
Fix usage of -->

flodiebold:
It might depend on the version of dash, but for me having the it in the nested
form did not work (it results in "Symbol's value as variable is void: it").


sebastiencs:
@flodiebold Thanks, just curious, what version of dash are you using ?

flodiebold:
@sebastiencs Apparently I'm using 20180206.2124, which seems to be the newest one available (from MELPA).

flodiebold:
25.3.1.

Issue #33, opened 2018-01-04
Conflicts with helm and solarized-theme


twlz0ne:
### Expected Behavior

helm CAN accept user input or keypresses

### Actual Behavior

helm CANNOT accept user input or keypresses

### Steps to reproduce the behavior

1. Download [gist: test-lsp-ui.el](https://gist.github.com/twlz0ne/75a7a54941a6ba54f66aaf4c5aeaba54)
1. `/path/to/emacs -Q -l /path/to/test-lsp-ui.el`, wait for the helm window to appear
1. Press any key

### Environment

- macOS 10.11.6
- Emacs 26 / 27
- Python 3.4.3 / 3.6.1
- lsp-ui-20180102.1122
- helm-20180102.2219
- solarized-theme-20180101.42

twlz0ne:
There are two solutions, but I do not understand its underlying causes.

1. Override the function `lsp-ui-peek--make-footer` in [lsp-ui/lsp-ui-peek.el#L204](https://github.com/emacs-lsp/lsp-ui/blob/b6523db5f428abaf247e7d5e84ed26e1335be284/lsp-ui-peek.el#L204):

    ```diff
        ;; delete line below or change `:height' to a number other than 0.5 to fix issue.
    -     (propertize "\n" 'face '(:height 0.5))
    ```

    Visit here to check out the context of modification: https://github.com/twlz0ne/lsp-ui/commit/2b86af41e25b7b3d2d1ef7f95994ccba6e1197d5.

2. Or change the difinition of `mode-line` in [solarized-theme/solarized.el#L298](https://github.com/bbatsov/solarized-emacs/blob/2dd2699b2f315374333292b132dc0dc03719aba2/solarized.el#L298):

    ```diff
        `(mode-line
        ((,class (:inverse-video unspecified
                                    :overline ,s-mode-line-bg
                                    :underline ,s-mode-line-underline
                                    :foreground ,s-mode-line-fg
                                    :background ,s-mode-line-bg
                                    ;; delete lines below or change the `:line-width' to \
                                    ;; a number other than 1 and 0 to fix issue.
    -                               :box (:line-width 1
    -                                                 :color ,s-mode-line-bg
    -                                                 :style unspecified)
                                    ))))
    ```

sebastiencs:
Wow that's a weird bug. Thanks for the detailed report !
I will look into it

sebastiencs:
```diff
[...]

PR #88, opened 2018-02-27
Resolves incompatibility with solarized theme and helm, and reduces slowdowns

arejensen:
This resolves the incompatibility issue with the solarized theme and helm (#33) and reduces slowdowns after the documentation frame is made invisible. Introduces new (other) slowdowns due to recreating the documentation frame instead of revealing it when documentation is shown.


MaskRay:
thx! i switched to ivy partially due to that issue...

PR #97, opened 2018-03-19
[lsp-ui-doc] Fix line breaks

twlz0ne:


sebastiencs:
Thanks, can you just rename `lsp-ui-doc--line-number-width` to `lsp-ui-doc--inline-line-number-width` ?

twlz0ne:
@sebastiencs updated

PR #102, opened 2018-03-20
permit `fringe-mode` to not be bound

### LINKED ISSUE #101: when fringe not available, weird errors
Maybe I have something misconfigured, but I run emacs from my console, and when using lsp-ui I get weird errors about `fringe-mode` being undefined and so forth. I *think* perhaps the fringe is not available without X or some similar system? Would be nice to detect this and gracefully back off.

nikomatsakis:
Fixes #101 -- but maybe not in the best way?


sebastiencs:
Yes that's fine. Thank you

PR #104, opened 2018-03-21
[lsp-ui-doc] Fix spaces calculation error

twlz0ne:
When using `length` to caculate Unicode string, We will get more space than we actually need
:

    (- MAX-WIDTH (length "中文")       DOC-STRING-WIDTH) ;; => (- MAX-WIDTH 2 DOC-STRING-WIDTH)
    (- MAX-WIDTH (string-width "中文") DOC-STRING-WIDTH) ;; => (- MAX-WIDTH 4 DOC-STRING-WIDTH)

After concatenated all (source code + spaces + doc string), the final width exceeds the window, line break happens:

screen_shot_2018-03-20_at_4_02_59_pm

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -461,16 +461,16 @@ Use because `string-width' counts invisible characters."
          (+ 2 (line-number-display-width))
        0)
      (if (bound-and-true-p linum-mode)
-         (cond ((stringp linum-format) linum-format)
+         (cond ((stringp linum-format) (string-width linum-format))
                ((eq linum-format 'dynamic)
-                (+ 2 (length (number-to-string
-                              (count-lines (point-min) (point-max)))))))
+                (+ 2 (string-width (number-to-string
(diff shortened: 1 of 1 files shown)


twlz0ne:
@sebastiencs Is it possible to replace all `(length ` with `(string-width ` to avoid potential problems?

twlz0ne:
@sebastiencs Yes. Using `string-width` solves the issue on my image.

According to my observation, this issue never appears in the Emacs GUI, because the `child-frame` is a floating window, Whether using `(length` or `(string-width`, or how wide the Unicode characters occupies.

sebastiencs:
Alright
> Is it possible to replace all (length with (string-width to avoid potential problems?

Do you need to replace `length` in others place or the PR is ready to be merged ?

twlz0ne:
@sebastiencs Just merge PR.
To replacing other `(length`, I need do more tests.

sebastiencs:
Alright. Thanks for working on this

PR #125, opened 2018-04-28
Flycheck report errors from all buffers in the workspace

yyoncho:


yyoncho:
@alexmurray thanks for the comment. I will change the code to make this behaviour configurable.

yyoncho:
ping

yyoncho:
Please ignore the pull request for now since there is performance issue related to find-buffer-visiting-file usage when there are many buffers.

MaskRay:
My elisp-fu sucks so I'm really uncomfortable to merge PRs.

@amosbird has a different diagnostics approach implemented as a pull mode.

yyoncho:
@amosbird can you share what you got? I am closing the PR since it would require more work in order to be merged.

amosbird:
Sorry for being late on this. I've added a lsp command that proactively request lsp-server to send current diagnostic information. Then all I need is `(add-hook 'lsp-after-diagnostics-hook #'flycheck-buffer)`

PR #126, opened 2018-04-29
Fix lsp-ui-sideline-apply-code-actions for multi-line titles

flodiebold:
Replace spaces and tabs just like on the sideline. Otherwise the completion doesn't work correctly, at least with helm.


MaskRay:
Do you have comparing screenshots?

brotzeit:
@flodiebold Is this still an issue ?

flodiebold:
I don't know, I haven't seen any multi-line code action titles in a while :thinking:

PR #127, opened 2018-05-01
[lsp-ui-doc] Backport string-trim changes from emacs 26.1

gpittarelli:
On the latest actual release of emacs (25.3), lsp-ui-doc master throws errors when trying to display docs. The cause is `(string-trim string "[\n]+")`: that second parameter will be added in 26.1. Not sure how to best resolve this; I did a straight backport of the functions from current Emacs master.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -33,6 +33,7 @@
 (require 'dash)
 (require 'dash-functional)
 (require 'markdown-mode)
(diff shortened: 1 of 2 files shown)


sebastiencs:
Thanks for you PR @gpittarelli, but I prefer another solution which doesn't involve backporting the function from emacs 26. I'm closing the PR, sorry

PR #128, opened 2018-05-01
[lsp-ui-doc] Add delay for showing docs

brettatoms:
Add a timer to delay showing the docs similar to the timer used in `lsp-ui-sideline.el`

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -57,6 +57,11 @@
   :type 'boolean
   :group 'lsp-ui-doc)
 
+(defcustom lsp-ui-doc-delay 0.2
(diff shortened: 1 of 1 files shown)


brettatoms:
There's an issue with this PR in that it takes the same amount of time for the doc to hide.  It should probably hide when the point is moved away from the symbol for the visible doc.

jsmestad:
Would love to see this added!

sebastiencs:
Implemented in https://github.com/emacs-lsp/lsp-ui/commit/3b91ef685ed7be5016d8c1b1bcdff177c4b7e75d @jsmestad.
Thanks @brettatoms, I had to change a few things

PR #131, opened 2018-05-09
Fix extraction of symbols in lsp-ui-sideline

kljohann:
As `forward-symbol` places point just _past_ the current symbol and not at the beginning of the next symbol some symbols were missed and several work-around were in place (e.g. checking for "::" or subtracting 1 from the column (even though this is already done by `lsp--cur-column`).

In particular for the following code
```
bool Something::check() { return true; }
```
the previous implementation would place point at the following positions:
```
bool Something::check() { return true; }
^   ^         ^      ^          ^    ^
```
where `Something` would be ignored because of the `(not (looking-at-p "::"))` check.
The new implementation looks at
```
bool Something::check() { return true; }
^    ^          ^         ^      ^
```
and considers all pointed-to symbols.

This pull request also introduces skipping of symbols in comments and checking for the hover capability.

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -386,45 +386,52 @@ to the language server."
   (lsp-ui-sideline--delete-ov)
   (when (and lsp--cur-workspace
              buffer-file-name)
-    (let* ((eol (line-end-position))
-           (eob (buffer-end 1))
-           (bol (line-beginning-position))
-           (doc-id (lsp--text-document-identifier))
-           (action-params
-            (if (equal lsp-ui-sideline-update-mode 'line)
-                (list :textDocument doc-id
-                      :range (lsp--region-to-range bol eol)
-                      :context (list :diagnostics (lsp--cur-line-diagnotics)))
-              (lsp--text-document-code-action-params)))
-           (line (point))
-           (line-widen (save-restriction (widen) (line-number-at-pos))))
+    (let ((eol (line-end-position))
+          (bol (line-beginning-position))
+          (line (line-number-at-pos))
+          (line-widen (save-excursion (widen) (line-number-at-pos)))
+          (doc-id (lsp--text-document-identifier)))
       (save-excursion
         (setq lsp-ui-sideline--occupied-lines nil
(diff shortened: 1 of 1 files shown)


kljohann:
@sebastiencs Sure, I just pushed a rebased version. Can you take a look?

lsp-ui-sideline.el
```diff
@@ -386,45 +386,51 @@ to the language server."
...
+              (let* ((symbol (thing-at-point 'symbol t))
+                     (bounds (bounds-of-thing-at-point 'symbol))
+                     (parsing-state (syntax-ppss))
+                     (in-string (nth 3 parsing-state))
+                     (outside-comment (eq (nth 4 parsing-state) nil)))
+                ;; Skip strings and comments
+                (when (and symbol (not in-string) outside-comment)
+                  (push (list symbol line bounds (lsp--cur-position)) symbols))))
+            (dolist (entry symbols)
+              (cl-destructuring-bind (symbol line bounds position) entry
```
sebastiencs: Since lsp-ui is using dash, can you use `-let` to destructure the variable ?

lsp-ui-sideline.el
```diff
@@ -386,45 +386,51 @@ to the language server."
...
+                        (progn (forward-symbol -1)
+                               (>= (point) bol)))
+              (let* ((symbol (thing-at-point 'symbol t))
+                     (bounds (bounds-of-thing-at-point 'symbol))
+                     (parsing-state (syntax-ppss))
+                     (in-string (nth 3 parsing-state))
+                     (outside-comment (eq (nth 4 parsing-state) nil)))
+                ;; Skip strings and comments
+                (when (and symbol (not in-string) outside-comment)
+                  (push (list symbol line bounds (lsp--cur-position)) symbols))))
```
sebastiencs: `lsp--cur-position` is using `line-number-at-pos` which can be very slow on big files.
Can you use the same logic that was done ? (Computing the current line before looping on every symbol and use it to send the position)

kljohann:
Done @sebastiencs .

sebastiencs:
Great. Thanks @kljohann !

PR #135, opened 2018-05-23
[lsp-ui-sideline] Update sideline information on position change

### LINKED ISSUE #111: Actions/flycheck notifications must be updated when text has changed
At this point, if you go and fix a flycheck notification it disappears from the flycheck list but it is still present in LSP UI (potential duplicate of https://github.com/emacs-lsp/lsp-ui/issues/106 ). The same is true for the actions. If you edit the text the actions are still present. Furthermore, you can still execute them over the new text and have an unpredictable result.

### LINKED ISSUE #121: Provide option to list errors from all buffers in the workspace in flycheck buffer
At this point, lsp-ui lists errors only for the current buffer(observable after invoking flycheck-error-list) and it has a separate method for listing all errors in the workspace. Since the latest version of flycheck supports displaying errors from multiple buffers it would be good to support that in lsp-ui as a option or as a default behaviour.

yyoncho:
At this point LSP-UI always sends region selection (current line) but some of
the servers (e. g. JDT) expect only the current point. The newly added option
will force lsp-ui to work like Visual Studio Code.

- Added option to switch from line to point mode - Fixes #121
- Triggered update when diagnostics changes - Fixes #111

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -71,6 +71,16 @@
   :type 'boolean
   :group 'lsp-ui-sideline)
 
+(defcustom lsp-ui-sideline-update-mode 'line
+  "Define the mode for updating sideline information.
+
+When set to `line' the information will be updated when user
+changes current line otherwise the information will be updated
+when user changes current point."
+  :type '(choice (const line)
+                 (const point))
+  :group 'lsp-ui-sideline)
+
 (defcustom lsp-ui-sideline-delay 0.2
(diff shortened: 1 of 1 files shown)


yyoncho:
@sebastiencs @MaskRay can someone review the change? I am working on LSP Java layer for Spacemacs and without this change, the most of the actions are not accessible.

sebastiencs:
Looks good, thanks. Sorry for the delay I don't have much free time theses days.

PR #138, opened 2018-05-30
Revert the update on diagnostics changed

yyoncho:
- Comment until the proper fix is found. At this point if there are 2 opened
files from the current workspace with diagnostics the second one will trigger
two subsecuqent sideline runs.


yyoncho:
I did some research and it looks like the issue is not caused by the flycheck change but it is only exposed by the change. Here are the steps to reproduce with original codebase:

1. Set `lsp-ui-sideline-delay` to 5 second
2. Go to buffer A and move the cursor so that lsp-ui-sideline will be triggered
3. Go to buffer in another window and do the same as in 2.
4. `lsp-sideline--run` runs two times in the second buffer.

yyoncho:
rebased.

sebastiencs:
Thanks @yyoncho !

PR #142, opened 2018-06-10
Don't request hover when server cannot handle textDocument/hover

wyuenho:
Related: https://github.com/emacs-lsp/lsp-mode/pull/375

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -274,7 +274,8 @@ We don't extract the string that `lps-line' is already displaying."
 (defun lsp-ui-doc--make-request ()
   "Request the documentation to the LS."
   (when (and (bound-and-true-p lsp--cur-workspace)
-             (not (bound-and-true-p lsp-ui-peek-mode)))
+             (not (bound-and-true-p lsp-ui-peek-mode))
+             (lsp--capability "hoverProvider"))
     (if (symbol-at-point)
         (let ((bounds (bounds-of-thing-at-point 'symbol)))
           (unless (equal lsp-ui-doc--bounds bounds)


gpittarelli:
I tested this with [groovy-language-server](https://github.com/palantir/language-servers) (which doesn't support hover) and it fixes an error :) Thanks

MaskRay:
In this case `lsp-ui-doc-mode` isn't useful. Can you just make `lsp-ui-doc-enable` a no-op?

wyuenho:
Following that logic, shouldn't all lsp-ui features be able to disable itself?

The way lsp-mode and lsp-ui are architected now, individual lsp-ui features know nothing about the server's state. They just naively register themselves to a bunch of standard emacs hooks, and attempt to make a request to a server, without regard whether it's up, initialized or ready.

Say I do disable lsp-ui-doc when textDocument/hover capability isn't found, the fact that capability isn't available doesn't tell me whether the server is being started, or initialized but not ready. So, when the server is started, initialized, and ready, I would have already disabled lsp-ui-doc, so what do I do now? Restart lsp-ui? The architecture and design in lsp-mode and lsp-ui is insane. If you are following a naive polling model, just stick with it instead of introducing yet another intermediate state that causes more trouble than it's worth. This is the simplest solution to solve the problem of "I'm suppose to get some documents from the server but I don't know anything about the server's state".

wyuenho:
@MaskRay Ping

MaskRay:
@wyuenho Sorry for taking so long to get to this. I am not good at elisp so not comfortable merging PRs even though I somehow get the permission. I was hoping to get more feedback from other more skilled people.

There seems a trend to make more stuff `*Provider`and the change is in line with recent lsp-mode changes so I'll merge this.

BTW, you might be interested at https://github.com/MaskRay/ccls/wiki/Getting-started but I'm very opinionated :)

Issue #140, opened 2018-06-06
lsp-ui-sideline breaks narrowing


Dickby:
with lsp-ui-sideline-mode enabled narrowing seems to not work anymore.
I have this problem by using lsp-ui with cquery.

muffinmad:
Broken in https://github.com/emacs-lsp/lsp-ui/commit/0d1c593d0a1ffbd446e10d3290221ed22f8ce257 by
```elisp
(line-widen (save-excursion (widen) (line-number-at-pos)))
```
It must be ```save-restriction``` instead of ```save-excursion``` to save narrowing

PR #145, opened 2018-06-14
Use save-restriction to save narrowing

muffinmad:
Fixes https://github.com/emacs-lsp/lsp-ui/issues/140


MaskRay:
Thx! Fixed in 62cc5c1166034be7880f0652eb1f19efabf4a657

PR #157, opened 2018-07-17
set left-fringe width to 0 for ui-doc

erikbye:
(left-fringe . (frame-char-width)) was probably used to create a margin for the text as to not mush it up against the frame. However, the margin created will use whatever background color the fringe face is set to, for its background color. Many have this set to black, so you end up with the black left-hand border.

Probably best to set left-fringe width to 0 until a better solution is in place for text padding.


MaskRay:
Can you give an example when this will give mismatched fringe color and background color?

```elisp
                           (left-fringe . ,(frame-char-width))
                           (background-color . ,(face-background 'lsp-ui-doc-background nil t))
```

Yanhao:
image

color theme is monokai, and with the default lsp-ui theme

PR #158, opened 2018-07-21
Fix lsp-ui-sideline-apply-code-actions for multi-line titles

flodiebold:
Replace spaces and tabs just like on the sideline.


flodiebold:
Sorry, didn't intend to create a duplicate.

PR #167, opened 2018-08-13
[lsp-ui-sideline] Do not mangle with flycheck when disabled

vincentbernat:
When lsp-ui-sideline-show-flycheck is set to nil, a user may expect
the default error handling (display errors in the echo area) to be
left untouched. However, lsp-ui disables flycheck error reporting.
This commit ensures this is done only when
lsp-ui-sideline-show-flycheck is set to t.


MaskRay:
Thanks

PR #183, opened 2018-09-30
[lsp-ui-doc] Wrap lines when not using childframes

nanzhong:
When not using childframes, lsp-ui-doc contents should be wrapped according to `lsp-ui-doc-max-width`.

Currently very long lines will result in a overlay that spans the entire width of the window that has truncated lines, eg.
afa52952-8bd0-48f8-9857-016dce435b5e
This is not helpful as the content is truncated and it also results in an inconsistent look.

Instead I think the overlay should be limited in width according to `lsp-ui-doc-max-wdith` resulting in the following:
d55c75a3-eaf2-4a00-a465-236229152c72

I’m quite new to elisp, so please let me know how I can improve my changes if needed. Thanks!

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -234,6 +234,12 @@ MODE is the mode used in the parent frame."
          (funcall render-fn string)
        (with-temp-buffer
          (insert string)
+
+         (if (lsp-ui-doc--inline-p)
+             (progn (let ((fill-column lsp-ui-doc-max-width))
+                      (fill-region (point-min) (point-max)))
+                    (buffer-substring (point-min) (point-max))))
+
          (delay-mode-hooks
            (let ((inhibit-message t))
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -234,6 +234,12 @@ MODE is the mode used in the parent frame."
          (funcall render-fn string)
        (with-temp-buffer
          (insert string)
+
+         (if (lsp-ui-doc--inline-p)
+             (progn (let ((fill-column lsp-ui-doc-max-width))
+                      (fill-region (point-min) (point-max)))
+                    (buffer-substring (point-min) (point-max))))
```
MaskRay: Why do you need `(buffer-substring (point-min) (point-max))`? The return value is not used.

I think you may use: when + fill-region

nanzhong:
@MaskRay good question. This was me pulling something from elsewhere for line wrapping strings. I’ve updated with your your suggestions.

PR #199, opened 2018-12-05
margin calculation should allow for end-of-line indicator

alanz:
In whitespace mode, the end of line marker adds am extra char.

This should probably be detected via the appropriate (if ..), but I have no idea how.

Closes https://github.com/emacs-lsp/lsp-mode/issues/499

My whitespace config is set to

```elisp
 '(whitespace-style
   (quote
    (face trailing tabs spaces lines-tail newline empty indentation space-after-tab space-before-tab tab-mark newline-mark))))
```

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -247,7 +247,8 @@ CURRENT is non-nil when the point is on the symbol."
          0)
      (if (bound-and-true-p display-line-numbers-mode)
          (+ 2 (line-number-display-width))
-       0)))
+       0)
+     1)) ;; AZ add one for the whitespace end-of-line marker
 
 (defun lsp-ui-sideline--window-width ()
   (- (min (window-text-width) (window-body-width))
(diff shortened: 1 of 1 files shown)


yyoncho:
Can you make this a variable like `lsp-ui-right-margin` defaulting to 0? I havent check that but guess it will have some minor visual implication when the user does not have that whitespace configuration.

alanz:
I actually think the calculation is correct, but the symbol display draws a box around the right-most word (the symbol name) and this pushes the emacs calculation over the edge. Because with this patch, visually there is a gap between the whitespace indicater `$` and the right margin.

screenshot from 2018-12-05 21-21-37

yyoncho:
Just tried it on my machine - it breaks when you enable whitespace-mode . The if you are looking could be `(bound-and-true-p whitespace-mode)`. We could additionally check whether whitespace-style contains newline-mark but it looks too much code for such a small problem.

yyoncho:
Thanks!

PR #208, opened 2018-12-14
Make it clear not to use flycheck-mode hooks for lsp

DougBeney:
This change makes it crystal clear that you should not use flycheck-mode hooks if you want syntax checking from the lsp.

For a while, I was so confused why my C++ includes were not find even though everything else was working. Turns out I misunderstood the readme. Hopefully this change helps anyone else who got confused by this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,9 @@ To enable the package and its features:
 (add-hook 'lsp-mode-hook 'lsp-ui-mode)
 ```
 
-To enable `flycheck-mode` for a particular LSP client, add the following
+**Note:** Be sure to remove previous `flycheck-mode` hooks you have in your config. They will remove lsp-mode capabilities.
+
+To use FlyCheck **without** lsp for a particular mode, add a hook like this.
 
 ```el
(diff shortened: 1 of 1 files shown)


yyoncho:
The readme is not up to date at all.  There is no need to do 
```
(require 'lsp-ui)
(add-hook 'lsp-mode-hook 'lsp-ui-mode)
```
or
```
(add-hook 'XXXXX-mode-hook 'flycheck-mode)
```

Since lsp-mode already handle this. If you want to use flycheck you only have to set lsp-prefer-flymake to nil and then call `lsp` and it will turn on flycheck/lsp-ui and so on.

DougBeney:
Ok, cool. That simplifies my config even more. Is help needed to update the readme or will it be updated for the next version release anyway?

yyoncho:
Any help is appreciated, it is Emacs community project, no matter whether it is documentation/new features/PRs/reviewing PRs/testing/etc .

brotzeit:
Done.

PR #211, opened 2018-12-16
LSP UI changes needed to make signature support work on lsp-ui side

### LINKED ISSUE #204: lsp-ui-doc--extract oddity for "MarkedString[]"
`lsp-ui-doc--extract`, used to render the pop-up window for hover text, has the following fragment for processing the `MarkedString[]` case

```elisp
   ((sequencep contents) ;; MarkedString[]
    (mapconcat 'lsp-ui-doc--extract-marked-string
               (lsp-ui-doc--filter-marked-string contents)
               "\n\n"
               ;; (propertize "\n\n" 'face '(:height 0.4))
               ))
```
This calls `lsp-ui-doc--filter-marked-string` on `contents`.  It it not clear to me what  `lsp-ui-doc--filter-marked-string` is intended to do, but it does invoke an idle timer function.  The nett effect for me is that the hover window only sometimes shows up.

Changing this to
```elisp
   ((sequencep contents) ;; MarkedString[]
    (mapconcat 'lsp-ui-doc--extract-marked-string
               contents
               "\n\n"
               ;; (propertize "\n\n" 'face '(:height 0.4))
               ))
```
Results in reliable display of the hover (except the very first time for a file, which I assume is an initialization issue, separate from this, see #178 )

So, is there any reason to call  `lsp-ui-doc--filter-marked-string` ?  If so, should it be a different version, without the idle timer randomly inserted in the middle?

yyoncho:
Fixes #204

Requires: https://github.com/emacs-lsp/lsp-mode/pull/544

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -251,9 +251,6 @@ MODE is the mode used in the parent frame."
   (let ((groups (--separate (and (hash-table-p it)
                                  (lsp-get-renderer (gethash "language" it)))
                             (append list-marked-string nil))))
-    (when-let ((marked-string (caar groups)))
-      ;; Without run-with-idle-timer, echo area will be cleared after displaying the message instantly.
-      (run-with-idle-timer 0 nil (lambda () (eldoc-message (lsp-ui-doc--extract-marked-string marked-string)))))
     (if lsp-ui-doc-include-signature
         list-marked-string
       (cadr groups))))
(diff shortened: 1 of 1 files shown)


alanz:
I just tried this version, the hover info shows up in the message area, but it is the raw markdown.

Is there a setting to get it to use the separate overlay window instead? as it used to

yyoncho:
Can you post a screenshot? I will install the haskell server so I can try it out.

alanz:
See

screenshot from 2018-12-16 18-39-43

Installing the haskell server is not for the faint of heart, but shout if you decide to do it and need help.

yyoncho:
I had some problems with the haskell ide(I had it working only in setup.hs) but the docs that were there looked like this:
emacs myoncho_072

Did you include https://github.com/emacs-lsp/lsp-mode/pull/544 when testing?

alanz:
nope, I brought in lsp-mode master. Will try again

alanz:
I brought that in, it made no difference until I disabled my prefer-flymake customisation.

Then I got 
screenshot from 2018-12-16 21-00-41

I think I need to back off my config, and build it back slowly, to see what is clashing.

But that will have to be tomorrow only.

yyoncho:
Ok, make sure that you have markdown-mode present since it is not a lsp.el dependency but without it you won't get markdown rendering.

PR #216, opened 2018-12-28
Fix disabling lsp-ui-doc-mode

iNecas:
The disable function was not removing the `lsp-ui-doc--on-hover` from the
`lsp-on-hover-hook` because of missing `local` flag.

Otherwise, the doc mode could not be disabled after enabling.


iNecas:
@yyoncho mind reviewing?

yyoncho:
Sorry for the delay, saw it over my phone and then forgot to merge it from my PC.

iNecas:
Sure thing. Happens to me all the time. Thanks

PR #221, opened 2019-01-07
[lsp-ui-doc] Render documentations in WebKit widgets

Jimx-:
This pull request provides an option to render documentations with a GTK+ WebKit widget in a child-frame.

This feature is enabled by `lsp-ui-doc-use-webkit` and requires Emacs to be built with the `--with-xwidgets` option.


lsp-ui-doc.el
```diff
@@ -229,8 +245,16 @@ MODE is the mode used in the parent frame."
                        (and (functionp lsp-ui-doc-render-function)
                             lsp-ui-doc-render-function)))
           (mode major-mode))
-     (if render-fn
-         (funcall render-fn string)
+     (cond
+      (lsp-ui-doc-use-webkit
+       (if (and language (not (string= "text" language)))
+           (progn
+             (message (format "```%s\n%s\n```" language string))
```
yyoncho: Is this message needed or it was only for debugging purpose?

lsp-ui-doc.el
```diff
@@ -86,6 +86,16 @@ Child frames requires GNU/Emacs version >= 26 and graphical frames."
...
   :group 'lsp-ui-doc)
 
+(defcustom lsp-ui-doc-use-webkit nil
+  "Whether to display documentation in a WebKit widget in a child-frame.
+This requires GNU/Emacs version >= 26 and built with the `--with-xwidgets`
+option."
+  :type 'boolean
+  :group 'lsp-ui-doc)
+
+(when lsp-ui-doc-use-webkit
```
yyoncho: This will work only if you set `lsp-ui-doc-use-webkit` upfront, but you could change this value afterwards. Can you move it just before using xwidget to render the doc?

lsp-ui-doc.el
```diff
@@ -229,8 +245,16 @@ MODE is the mode used in the parent frame."
                        (and (functionp lsp-ui-doc-render-function)
                             lsp-ui-doc-render-function)))
           (mode major-mode))
-     (if render-fn
-         (funcall render-fn string)
+     (cond
+      (lsp-ui-doc-use-webkit
```
yyoncho: You should add also (featurep 'webkit).

yyoncho:
Just one more comment: in order to make this work we need to modify the melpa recipe to have the html packaged with lsp-ui.

Jimx-:
Done! It disables webkit and uses normal child-frame when xwidget support is not available.

Cask
```diff
@@ -7,4 +7,4 @@
 (depends-on "markdown-mode")
 
 (package-file "lsp-ui.el")
-(files "*.el")
+(files "*.el" "lsp-ui-doc.html")
```
yyoncho: I am not sure whether this is sufficient but I suspect that melpa does not take this into account but uses only this file https://github.com/melpa/melpa/blob/master/recipes/lsp-ui

Compro-Prasad:
How can I enable this?

Jimx-:
Make sure that your Emacs is compiled with xwidget support (please refer to https://jiegec.me/programming/2016/02/18/building-emacs-git-version-with-xwidgets-and-modules-in-archlinux/ for more details). After that, add the following lines to your init.el:
```
(setq lsp-ui-doc-use-webkit t)
(require 'lsp-ui)
```

Compro-Prasad:
I don't know why but I am still getting eldoc only.

Compro-Prasad:
Okay I could finally connect to the Emacs China server and have installed @Jimx- Emacs config. But it doesn't enable this feature. Can you please share what else should I do to enable this feature?

yyoncho:
I believe that you need:

1. Checkout this PR
2. Verify that xwidgets works - you could do that by calling `xwidget-webkit-browse-url`
3. Set  `lsp-ui-doc-use-webkit` to t
4. Enable lsp-ui-mode in the buffer (or `lsp-ui-doc-mode`)

Jimx-:
If you are using `use-package` then you may try this configuration. Please replace the load-path by your local repo with this PR
```emacs-lisp
(use-package lsp-ui
  :load-path "~/projects/lsp-ui"
  :hook (lsp-mode . lsp-ui-mode)
  :init
  (setq lsp-ui-doc-use-webkit t)
  :config
  (setq lsp-ui-doc-enable t
        lsp-enable-completion-at-point t
        lsp-ui-doc-position 'at-point
        lsp-ui-doc-header nil
        lsp-ui-doc-include-signature t
        lsp-ui-sideline-enable nil))
```

Compro-Prasad:
I am sorry for not using a xwidgets enabled Emacs but it still doesn't work. I am using your configuration. I think it isn't able to enable itself.

I am testing this using ccls on a C++ project. Value of `lsp-ui-doc-use-webkit` is `t` in the buffer I am testing in.

I could test this if you could provide me with some reference on how to run it using `M-:`.

yyoncho:
@Compro-Prasad you may rebase to include the following commit: https://github.com/emacs-lsp/lsp-ui/commit/f9ff951ac956b7fc57bd63788cd12c896ab75948 

It adds interactive commands for displaying the doc.

MaskRay:
My window manager is xmonad. The xwidget frontend renders slowly and the appearance doesn't look great. Do you have idea how to improve its appearance?

Jimx-:
@Compro-Prasad What's the version of Emacs that you use? Does lsp-ui display the doc in a normal child-frame, or it displays nothing at all? The webkit frontend requires child-frame feature which is only available on Emacs version >= 26.

Jimx-:
@MaskRay Because webkit executes scripts asynchronously, the frame is resized before rendering finishes which causes the the frame to flicker. This should be fixed by the last commit.

Also, I use this set of frame parameters to remove the left border for a better appearance:
```emacs-lisp
(setq-default lsp-ui-doc-frame-parameters
                '((left . -1)
                  (top . -1)
                  (no-accept-focus . t)
                  (min-width . 0)
                  (width . 0)
                  (min-height . 0)
                  (height . 0)
                  (internal-border-width . 0)
                  (vertical-scroll-bars)
                  (horizontal-scroll-bars)
                  (left-fringe . 0)
                  (right-fringe . 0)
                  (menu-bar-lines . 0)
                  (tool-bar-lines . 0)
                  (line-spacing . 0.1)
                  (unsplittable . t)
                  (undecorated . t)
                  (minibuffer . nil)
                  (visibility . nil)
                  (mouse-wheel-frame . nil)
                  (no-other-frame . t)
                  (cursor-type)
                  (no-special-glyphs . t)))
```

Compro-Prasad:
@Jimx- Using latest Emacs build. I think I need to nuke my configuration to get this working. Thanks for your help though.

Compro-Prasad:
@yyoncho Thanks for the commit. Will check if it works.

Compro-Prasad:
I ran `M-x lsp-ui-doc-show` and it doesn't do anything for me. Then I did `M-: (lsp-ui-doc-show)` which returned a frame object and nothing poped up.

Compro-Prasad:
Okay, it popped up. I tried toggling `lsp-ui-mode` and it somehow worked.

Compro-Prasad:
screenshot from 2019-01-14 12-08-50

Here are some of my thoughts:

1. The width seems constant. On some xwidget popups a horizontal scroll bar is shown even though the window size is wide enough to fit the full length of the line in the documentation. It would be great if the width auto adjusts according to both the documentation and the window width/height. Exposing a variable to adjust the size would also be super useful.

2. My laptop isn't a beast, so the first documentation popup takes lot of time(10secs-1min). Can this be somehow eliminated?

3. I also get the same margin as https://github.com/emacs-lsp/lsp-ui/pull/221#issuecomment-453826346 . Sometimes it is on both the sides. It isn't a big issue but that margin doesn't seem to be useful to me.

Jimx-:
1. The width/height are actually readjusted according to the size of contents. You can move the cursor to a symbol with longer documentation to see that. The problem is the size of the container is decided by the browser which renders the documentation so it's not easy to control that from the elisp side. I simply set the min/max width of the container manually and the browser calculates the height based on that. Also, the Github markdown stylesheet prefers to render some elements(e.g. code blocks) in a single line with a scroll bar instead of breaking the line to fit the width. This can be fixed by tuning the stylesheet though.

2. The first documentation will not pop up sometimes because lsp-ui attempts to render the content before the browser finishes loading the page. You can move your cursor to somewhere else and then back to the first symbol to reset the rendering and see if it pops up immediately.

3. The margin is a result of your frame parameters and becomes noticeable as the buffer and the browser use different background colors. You can use a different set of `lsp-ui-doc-frame-parameters` to eliminate the margin.

sebastiencs:
That's very nice ! A few notes:

- On my high dpi display, the font is very small
I'm not sure if it's a bug in emacs or if it should be handled with some lisp code.

- When the documentation change, there are some flickering issue, it's not really visible on this gif but it's clear on the last frames:

peek 14-01-2019 21-37

- Scrolling on the frame has no effect, but if I scroll it and then display another documentation, there is still the frame but without the documentation in it

lsp-ui-doc.el
```diff
@@ -274,13 +296,66 @@ We don't extract the string that `lps-line' is already displaying."
...
+     (insert " ")
+     (goto-char 1)
+     (let ((id (make-xwidget
+                'webkit
+                nil
+                1
+                1
+                nil
+                (buffer-name))))
+       (xwidget-resize id 1 1)
```
sebastiencs: Is this call to `xwidget-resize` necessary ? I see you're already initializing the widget with the same size

lsp-ui-doc.el
```diff
@@ -331,7 +415,9 @@ START-X is the position x of the current window.
 START-Y is the position y of the current window."
   (-let* (((x . y) (--> (bounds-of-thing-at-point 'symbol)
                         (nth 2 (posn-at-point (car it)))))
-          (mode-line-y (lsp-ui-doc--line-height 'mode-line))
+          (mode-line-y (if lsp-ui-doc-use-webkit
+                           0
+                         (lsp-ui-doc--line-height 'mode-line)))
```
sebastiencs: Why change this with webkit only ? I think it should keep the same behavior than native frame. no ?

sebastiencs:
See the following diff, it fixes scaling with high dpi display

```patch
diff --git a/lsp-ui-doc.el b/lsp-ui-doc.el
index 5e5547c..9598cdb 100644
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -163,6 +163,13 @@ The functions receive 2 parameters: the frame and its window.")
           "lsp-ui-doc.html")
   "Path to the page loaded when a WebKit widget is created.")
 
+(defvar lsp-ui-doc--webkit-scale
+  (string-to-number (or (getenv "GDK_SCALE") "1"))
+  "Scale the widget.  It should be an integer according to the gtk doc:
+https://developer.gnome.org/gtk3/stable/gtk-x11.html.
+
+By default, its value is the environment variable GDK_SCALE or 1 if unset.")
+
 ;; Avoid warning with emacs < 26
 (declare-function display-buffer-in-child-frame "window.el")
 
@@ -384,7 +391,10 @@ We don't extract the string that `lps-line' is already displaying."
       (lsp-ui-doc--line-height)))
 
 (defun lsp-ui-doc--webkit-resize-callback (size)
-  (xwidget-resize (lsp-ui-doc--webkit-get-xwidget) (aref size 0) (aref size 1))
+  (xwidget-resize (lsp-ui-doc--webkit-get-xwidget)
+                  ;; + 20 for css padding
+                  (round (+ 20 (* lsp-ui-doc--webkit-scale (aref size 0))))
+                  (round (+ 20 (* lsp-ui-doc--webkit-scale (aref size 1)))))
   (lsp-ui-doc--move-frame (lsp-ui-doc--get-frame)))
 
 (defun lsp-ui-doc--resize-buffer ()
diff --git a/lsp-ui-doc.html b/lsp-ui-doc.html
index 1895ce7..279cdeb 100644
--- a/lsp-ui-doc.html
+++ b/lsp-ui-doc.html
@@ -11,7 +11,7 @@
        box-sizing: border-box;
        min-width: 400px;
        max-width: 600px;
-       margin: 0 auto;
+       margin: 0;
        padding: 10px;
      }
     </style>
```

Jimx-:
@sebastiencs I noticed that if we add the css padding to the widget size the frame will become wider every time the documentation pops up.

yyoncho:
https://github.com/melpa/melpa/pull/5981 to update melpa.

MaskRay:
`(setq lsp-ui-doc-position 'at-point)` does not work. Stack trace:

```elisp
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  /(nil 2)
  (> y (/ mode-line-y 2))
  (and (> y (/ mode-line-y 2)) (<= (- mode-line-y y) (+ char-height height)) (> (- y height) 0) (- y height))
  (or (and (> y (/ mode-line-y 2)) (<= (- mode-line-y y) (+ char-height height)) (> (- y height) 0) (- y height)) (+ y char-height))
  (let* ((--dash-source-64-- (let ((it (bounds-of-thing-at-point 'symbol))) (nth 2 (posn-at-point (car it))))) (x (car-safe (prog1 --dash-source-64-- (setq --dash-source-64-- (cdr --dash-source-64--))))) (y --dash-source-64--) (mode-line-y (lsp-ui-doc--line-height 'mode-line)) (char-height (frame-char-height)) (y (or (and (> y (/ mode-line-y 2)) (<= (- mode-line-y y) (+ char-height height)) (> (- y height) 0) (- y height)) (+ y char-height)))) (set-frame-position frame (+ x start-x) (+ y start-y)))
  lsp-ui-doc--mv-at-point(#<frame  *lsp-ui-doc-67109191* – Doom Emacs 0x557703e87040> 72 0 0)
  (if (eq lsp-ui-doc-position 'at-point) (lsp-ui-doc--mv-at-point frame height left top) (set-frame-position frame (if (and (>= left (+ width 10 (frame-char-width))) (not (lsp-ui-doc--next-to-side-window-p))) 10 (- (frame-pixel-width) width 10 (frame-char-width))) (cond ((eq lsp-ui-doc-position 'top) (+ top 10)) ((eq lsp-ui-doc-position 'bottom) (- (lsp-ui-doc--line-height 'mode-line) height 10)) (t nil))))
  (let* ((--dash-source-65-- (window-edges nil nil nil t)) (left (car-safe (prog1 --dash-source-65-- (setq --dash-source-65-- (cdr --dash-source-65--))))) (top (car-safe (prog1 --dash-source-65-- (setq --dash-source-65-- (cdr --dash-source-65--))))) (_bottom (car (cdr --dash-source-65--))) (window (frame-root-window frame)) (--dash-source-66-- (window-text-pixel-size window nil nil 10000 10000 t)) (width (car-safe (prog1 --dash-source-66-- (setq --dash-source-66-- (cdr --dash-source-66--))))) (height --dash-source-66--) (width (+ width (* (frame-char-width frame) 1))) (char-h (frame-char-height)) (height (min (- (* lsp-ui-doc-max-height char-h) (/ char-h 2)) height)) (frame-resize-pixelwise t)) (set-frame-size frame width height t) (if (eq lsp-ui-doc-position 'at-point) (lsp-ui-doc--mv-at-point frame height left top) (set-frame-position frame (if (and (>= left (+ width 10 (frame-char-width))) (not (lsp-ui-doc--next-to-side-window-p))) 10 (- (frame-pixel-width) width 10 (frame-char-width))) (cond ((eq lsp-ui-doc-position 'top) (+ top 10)) ((eq lsp-ui-doc-position 'bottom) (- (lsp-ui-doc--line-height 'mode-line) height 10)) (t nil)))))
  lsp-ui-doc--move-frame(#<frame  *lsp-ui-doc-67109191* – Doom Emacs 0x557703e87040>)
  lsp-ui-doc--webkit-resize-callback([600 72])
```

Jimx-:
Looks like a WM specific issue. I got the same error with KWin but not with I3. Providing a default mode line height should fix it.

yyoncho:
The melpa PR was merged, @sebastiencs and @MaskRay do you have more comments?

sebastiencs:
The name of the function `lsp-ui-doc--line-height` is badly chosen.
The function returns the position y of the line (the docstring is correct):

```elisp
(defun lsp-ui-doc--line-height (&optional line)
  "Return the pos-y of the LINE on screen, in pixel."
  (or (nth 2 (or (window-line-height line)
                 (and (redisplay t)
                      (window-line-height line))))
      ;; If window-line-height returns nil, calculate the position
      ;; y of the modeline ourself, this will give the right value
      ;; only if the modeline has the same height that a normal
      ;; frame character
      (and (eq line 'mode-line)
           (- (window-pixel-height) (frame-char-height)))
      0))
```

sebastiencs:
The feature is nice but in practice I experience lots of flickering (see @MaskRay's comment and mine)

Is there a system/window manager where there isn't this behavior ? If yes, I think it should be good to indicate them on the readme and in the docstring of `lsp-ui-doc-use-webkit`

In all cases, someone should probably report this to `bug-gnu-emacs@gnu.org`, they can for sure help us to reduce/eliminate the flickering issue and improve the support for hdpi screens.

mpereira:
A bit off-topic, but was anyone able to compile Emacs with xwidgets support on MacOS? After jumping through a few hoops, I'm currently having trouble building webkitgtk+, which is required when building emacs with xwidgets.

shackra:
This does not work smooth for me for some reason, here is a video:
https://youtu.be/ohAC6xSfNE8
This is my lsp-* configuration:
```
(use-package lsp-clients
  :straight lsp-mode
  :config
  (add-hook 'lsp-after-open-hook 'lsp-enable-imenu)
  :init
  (setf lsp-eldoc-render-all nil)
  (setq lsp-inhibit-message t)
  (setq lsp-message-project-root-warning t)
  (setf lsp-prefer-flymake nil)
  ;; arreglo rápido; hay que correr `yarn global add
  ;; typescript-language-server'
  ;; https://github.com/emacs-lsp/lsp-mode/issues/588
  (setq lsp-clients-typescript-server "typescript-language-server"
        lsp-clients-typescript-server-args '("--stdio")))

(use-package lsp-ui
  :after (lsp)
  :init
  (setf lsp-ui-sideline-enable nil)
  (when (require 'xwidget nil 'noerror)
    (setf lsp-ui-doc-use-webkit t))
  (when lsp-ui-doc-use-webkit
    (setf lsp-ui-doc-enable t)
    (setf lsp-ui-doc-position 'at-point)
    (setf lsp-ui-doc-header nil)
    (setf lsp-ui-doc-include-signature t))
  :config
  (define-key lsp-ui-mode-map [remap xref-find-definitions] #'lsp-ui-peek-find-definitions)
  (define-key lsp-ui-mode-map [remap xref-find-references] #'lsp-ui-peek-find-references)
  (add-hook 'lsp-after-open-hook #'lsp-ui-mode))

(use-package company-lsp)
```

Narice:
I posted an issue on their board.
Would it be possible for you @Jimx- to explain the mentionned section of code in the issue I posted or here?
It seems to me that the language server might be doing something odd.
Being the only place I see where markdown quotes are added, my guess would be that the elm-language-server is triggering this region of code for some reasons.
Don't hesitate to tell me if this is not accurate (I don't know any elisp so this is just speculation).

Jimx-:
@Narice Could you try if this commit https://github.com/Jimx-/lsp-ui/commit/af7279c24ae396e1eeee4485bdc324b68c237fdf fixes the issue? Normally these quotes are added only if the contents are tagged with a specific language other than markdown. However it seems that the latest lsp-ui sets language to markdown if there is no language tag so the markdown gets wrapped inside another code block.

Narice:
oh I see, so I guess the language is elm instead of markdown or something like that?
I'll try out your commit as soon as possible.

Narice:
I redefined the function in my config.el and it works fine!
So your modification works ;)
there seem to be problem other than than in the rendering:
image
image
the last one updates to that when scrolled:
image
But this is another issue than the one you resolved hehe

Jimx-:
I cannot reproduce it. Could you try cloning the whole repo and load it instead of patching a single function in your config.el? It looks like the frame had the correct size before you made the modification.

Issue #150, opened 2018-06-21
lsp-ui-sideline only show the first line of error.


nasyxx:
I am not sure if it is a bug of the sideline or my wrong config.

The error shows in normal:

<img width="455" alt="screen shot 2018-06-21 at 7 47 47 pm" src="https://user-images.githubusercontent.com/5007613/41717769-6662b6a2-758d-11e8-8f9e-fa64c4141a0c.png">

The error shows in sideline:
<img width="650" alt="screen shot 2018-06-21 at 7 48 15 pm" src="https://user-images.githubusercontent.com/5007613/41717774-69e72826-758d-11e8-99ac-312032f5d000.png">

whirm:
I'm hitting the same issue with `ocaml-langague-server`.

thirdeyefunction:
@coreyconnor I tried your changes with ocaml-language-server and while it's an improvement, the alignment is kind of wonky: ![Imgur Image](https://i.imgur.com/I8GYBY0.png)

coreyoconnor:
@thirdeyefunction I think there are two issues there: The first error message line has extra spaces at the end; the use of right justification for the block is questionable.

The first issue should be easy: always trim spaces from each line. I think that's universally applicable.

As for the second, the right justification, not sure if there is a universal treatment. I'm fine with the right justification. In this case tho, with the line being so short and the added spaces, it really stands out.

coreyoconnor:
This is complete. All recent changes are only rebases. Please merge.

aqxa1:
I just want to note that this issue is back as of 401e69bffc9d581fb979389fdf6de47c3f4084dd, since `lsp-ui-sideline-diagnostic-max-lines` is now set to `1`. Fair enough if the old default causes issues, but if anyone is wondering why it's broken again, that's why. Easy to fix on the config side, thankfully:
`(setq lsp-ui-sideline-diagnostic-max-lines 3)`
or set through the customize interface.

PR #229, opened 2019-01-26
show all lines of an error. #150

coreyoconnor:
This has an issue where the error line order might output incorrectly. Not skilled in emacs/elisp so any suggestions would be great :)

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -32,6 +32,8 @@
 (require 'lsp-mode)
 (require 'flycheck nil 'noerror)
 (require 'dash)
+(require 'seq)
+(require 'subr-x)
 
 (defgroup lsp-ui-sideline nil
   "Display informations of the current line."
@@ -84,6 +86,11 @@ when user changes current point."
 (defcustom lsp-ui-sideline-delay 0.2
   "Number of seconds to wait before showing sideline."
   :type 'number
+  :group 'lsp-ui-sideline)
+
+(defcustom lsp-ui-sideline-diagnostic-max-lines 20
+  "Maximum number of lines to show of diagnostics in sideline."
+  :type 'integer
   :group 'lsp-ui-sideline)
 
 (defvar lsp-ui-sideline-code-actions-prefix ""
@@ -159,16 +166,18 @@ INDEX is the line number (relative to the current line)."
         (when (>= (- win-width (current-column)) str-len)
           eol)))))
 
-(defun lsp-ui-sideline--find-line (str-len bol eol &optional up)
+(defun lsp-ui-sideline--find-line (str-len bol eol &optional up offset)
   "Find a line where the string can be inserted.
 It loops on the nexts lines to find enough space.
 Returns the point of the last character on the line.
 
 STR-LEN is the string size.
 BOL & EOL are beginning and ending of the user point line.
-if UP is non-nil, it loops on the previous lines.."
+if UP is non-nil, it loops on the previous lines.
+if OFFSET is non-nil, it starts search OFFSET lines from user point line."
   (let ((win-width (lsp-ui-sideline--window-width))
-        (index 1) pos)
+        (index (if (null offset) 1 (1+ offset)))
+        pos)
     (while (and (null pos) (<= (abs index) 30))
       (setq index (if up (1- index) (1+ index)))
       (setq pos (lsp-ui-sideline--calc-space win-width str-len index)))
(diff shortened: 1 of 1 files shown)


coreyoconnor:
Here is a contrived example of a multi-line error message displayed using this change:

image

Without this change only the "scalac: missing argument list for method run in class IndexedReaderWriterStateT" text is displayed.

sebastiencs:
Thanks @coreyoconnor, could you add a `defcustom` to enable this behavior, as suggested by @yyoncho.

The issue with this change is that it could display lines of an error in disorder.
`lsp-ui-sideline--find-line` will return the first line (position) which has enough space to display that line.

For instance, if an error is:
```
..........First line of an error..........
Second line
```

It could be display like this:
```
Second line
..........First line of an error..........
```

To fix this you need to check that `lsp-ui-sideline--find-line` returns always a number (position) superior to the one in the previous iteration

nasyxx:
So glad that this issue can be solved.  Thank you very much. 

For a long time, when I used Haskell, I closed the sideline's flycheck because it displays too few and even only one "error".

coreyoconnor:
@nasyxx have you tried these changes with Haskell? I've only tried Scala. As @sebastiencs noted there is an issue with ordering but otherwise should work.

Should be easy to fix... but my elisp time is up for a bit :)

nasyxx:
I have tried it; it did the fix.

screen shot 2019-01-30 at 8 13 37 am

The old one only has the message which inside `[]`.

coreyoconnor:
I added a defcustom. I won't be correcting the out-of-order aspect in the foreseeable future. (non critical)

The option `lsp-ui-sideline-diagnostic-all-lines` defaults to true. I don't know of a usecase where only showing the first line is useful. So that's reasonable to me.

That said, I think a  "maximum number of lines" seems a better option. What do you think? Default to something large. Would cover the single line case, the usual, and mitigate those servers that output too much. (Not sure which servers do that but I suspect C++ would haha)

Also, I'd prefer if this was per-mode but I'm unsure how to do that. defvar-local?

lsp-ui-sideline.el
```diff
@@ -305,23 +312,33 @@ CURRENT is non-nil when the point is on the symbol."
...
+                   (message (progn (add-face-text-property 0 len 'lsp-ui-sideline-global nil message)
+                                   (add-face-text-property 0 len face nil message)
+                                   message))
+                   (string (concat (propertize " " 'display `(space :align-to (- right-fringe ,(lsp-ui-sideline--align len margin))))
+                                   message))
+                   (pos-ov (lsp-ui-sideline--find-line len nil))
+                   (ov (and pos-ov (make-overlay pos-ov pos-ov))))
+              (when pos-ov
+                (overlay-put ov 'after-string string)
+                (push ov lsp-ui-sideline--ovs)))
```
yyoncho: please place these on one line

yyoncho: (I mean the brackets)

coreyoconnor:
Will do!

lsp-ui-sideline.el
coreyoconnor: Is there a reference for the style? I think I corrected this.

coreyoconnor:
@yyoncho Can you review again? Instead of a boolean the change now introduces a `lsp-ui-sideline-diagnostic-max-lines` integer variable.

lsp-ui-sideline.el
```diff
@@ -306,24 +313,35 @@ CURRENT is non-nil when the point is on the symbol."
...
+                   (string (concat (propertize " " 'display `(space :align-to (- right-fringe ,(lsp-ui-sideline--align len margin))))
+                                   message))
+                   (pos-ov (lsp-ui-sideline--find-line len nil))
+                   (ov (and pos-ov (make-overlay pos-ov pos-ov)))
+                   )
+              (when pos-ov
+                (overlay-put ov 'after-string string)
+                (overlay-put ov 'kind 'diagnotics)
+                (push ov lsp-ui-sideline--ovs)
+                )
```
yyoncho: Can you put all the closing brackets on one line?

coreyoconnor: Done. I think? I'm ignorant of emacs style and style tooling.

coreyoconnor:
@sebastiencs 

> To fix this you need to check that `lsp-ui-sideline--find-line` returns always a number (position) superior to the one in the previous iteration

Completed a first pass at this. `lsp-ui-sideline--find-line` takes an optional parameter `offset` that is incremented each iteration. The function now returns cons `pos` and `index` which is used to update offset. Please review. :)

coreyoconnor:
This is complete. All recent changes are rebases. Please merge :)

PR #233, opened 2019-01-30
[lsp-ui-doc] Fix deleting the child frame if window changes (#224)

### LINKED ISSUE #224: Doc child frame should disappear when moving between windows
image

The doc displayed on the right actually belongs to the left window. This is fine since I set `lsp-ui-doc-use-childframe` and I'd like to keep it. However, when moving to the right Magit window the doc child frame stays open which hides a part of the Magit buffer.

I noticed that there's an advise around `select-window` which should handle this case, but it doesn't seem to work for me. I use windmove to change windows, but also using `other-window` doesn't make any different.

aclonegeek:
I'm new to using elisp, so hopefully this is the right way to go about fixing this one!

The defadvice for handling this case currently doesn't work because the condition always evaluates to true when selecting windows (i.e. ad-get-arg 0 and selected-window are always equal) - it looks like the selected window is being compared with itself.

I believe the expected behaviour should be to delete the child frame if a window is selected in general, so a check would be unnecessary.

Fixes #224

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -557,8 +557,7 @@ HEIGHT is the documentation number of lines."
 
 (defadvice select-window (after lsp-ui-doc--select-window activate)
   "Delete the child frame if window changes."
-  (unless (equal (ad-get-arg 0) (selected-window))
-    (lsp-ui-doc--hide-frame)))
+  (lsp-ui-doc--hide-frame))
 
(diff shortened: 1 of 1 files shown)


antifuchs:
I agree that this might not be the right fix (and that #89 isn't quite right). I made #250 to fix the logic and preserve the doc frame if it doesn't need hiding.

yyoncho:
Closing in favour of #250

Issue #234, opened 2019-01-30
 lsp-ui-flycheck-enable seems to be ignored


Fuco1:
There's a `;; FIXME: Provide a way to disable lsp-ui-flycheck` in the code.  Wouldn't it be enough to wrap this function with the defcustom variable we already have?

I don't want to use flycheck from lsp as for many modes there are better tools.

yyoncho:
The only option to make this happen at this point as I commented in the other PR is to set `lsp-auto-configure` to nil and then write your custom configuration code (you may use `lsp--auto-configure` as an example). Please comment on whether this suits you.

> I don't want to use flycheck from lsp as for many modes there are better tools.

Just curious, which modes/language servers you are referring to?

Fuco1:
I did this simple change and it solves my problem

```
;; FIXME: Provide a way to disable lsp-ui-flycheck
(defun lsp-ui-flycheck-enable (_)
  "Enable flycheck integration for the current buffer."
  (when lsp-ui-flycheck-enable          ;;;;;;;;; THIS CONDITION WAS ADDED
    (when lsp-ui-flycheck-live-reporting
      (setq-local flycheck-check-syntax-automatically nil))
    (setq-local flycheck-checker 'lsp-ui)
    (lsp-ui-flycheck-add-mode major-mode)
    (add-to-list 'flycheck-checkers 'lsp-ui)
    (add-hook 'lsp-after-diagnostics-hook 'lsp-ui-flycheck--report nil t)))
```

Is this something we forgot to put it? There's the FIXME in the code already.

Fuco1:
I'm using https://github.com/phpstan/phpstan for php and flow for javascript which are much more advanced than the php lsp (which checks syntax only).

yyoncho:
There is PR at https://github.com/emacs-lsp/lsp-ui/pull/235

The other option is to introduce one more terminal value for `lsp-prefer-flymake`  `:none` which will mean - do not enable anything.

yyoncho:
@Fuco1 I have pushed https://github.com/emacs-lsp/lsp-mode/commit/a8e549d92e67dd860e471224bd60654b22ca2b8f . Although what you have proposed will work too it does not make sense to have a defcustom which will practically skip the body of a function when called. It makes more sense to have a setting which skips the function call.

PR #235, opened 2019-01-30
Add variable for disabling lsp-ui flycheck checker

jvalkeejarvi:


yyoncho:
The issue was fixed in lsp-mode - check https://github.com/emacs-lsp/lsp-ui/issues/234#issuecomment-459491638

PR #249, opened 2019-02-28
lsp-ui-doc: fix handling of markdown MarkupContent

Khady:
```
>>> ocamlmerlin-lsp:543(async)
Content-Length: 218

{
  "jsonrpc": "2.0",
  "method": "textDocument/hover",
  "params": {
    "textDocument": {
      "uri": "file:///tmp/lsp/test.ml"
    },
    "position": {
      "line": 6,
      "character": 23
    }
  },
  "id": 3
}

<<<< ocamlmerlin-lsp:543
{
  "id": 3,
  "jsonrpc": "2.0",
  "result": {
    "contents": {
      "value": "```ocaml\nint\n```",
      "kind": "markdown"
    }
  }
}
```

Before this patch, the value in the lsp-server response would be returned as plaintext. It is now rendered as markdown.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -233,14 +233,15 @@ MODE is the mode used in the parent frame."
              (split-string string "\n")
              "\n"))
 
-(defun lsp-ui-doc--extract-marked-string (marked-string)
+(defun lsp-ui-doc--extract-marked-string (marked-string &optional language)
   "Render the MARKED-STRING."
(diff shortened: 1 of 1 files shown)


Khady:
This fix is not totally complete. The webkit rendering is not good. It surrounds the value with backticks. But this is not correct as the spec says the value can be a whole markdown document, not just a block of code. I don't have a good setup to try tho.

PR #250, opened 2019-03-01
Fix same-window logic in select-window advice for doc frame

antifuchs:
This PR is a fix for #224, preserving (what I believe is) the intent behind the change in #89: To hide the doc view only if the currently-selected window is no longer the one that the doc view is attached to.

The [original defadvice](https://github.com/emacs-lsp/lsp-ui/blob/df3681b6a578a314374a8d689675aa60ad4afa6e/lsp-ui-doc.el#L670) from #89 was subtly not-quite-right (causing #224) by being an `'after` advice: The advice checked whether `(selected-window)` was the same as the one selected, which _after_ `select-window` had run, was always the case: the `unless` never triggered, always preventing the frame from hiding.

With this PR, I hope to change the check so it is more meaningful, by making the advice an `'around` class, and checking what the values of the selected window are before and after `select-window` ran.

Further, this change prevents the doc frame from hiding if the newly-selected window is the minibuffer (which might be nice for things like `go-rename` and other tools that query for things, where someone might want to use documentation as a reference).

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -678,9 +678,12 @@ before, or if the new window is the minibuffer."
   (let ((initial-window (selected-window)))
     (prog1 ad-do-it
       (when (lsp-ui-doc--visible-p)
-        (let ((current-window (selected-window)))
+        (let* ((current-window (selected-window))
+               (doc-buffer (get-buffer (lsp-ui-doc--make-buffer-name))))
           (unless (or (window-minibuffer-p current-window)
-                      (equal current-window initial-window))
+                      (equal current-window initial-window)
+                      (and doc-buffer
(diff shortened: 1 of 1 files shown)


antifuchs:
...huh, yes, I see it too now. I'll find out what is causing this & revise the patch.

antifuchs:
Found the problem: When popping up the documentation buffer's frame, that frame's window would briefly be selected, and so since lsp-ui-doc selects the original frame back, after doc frame creation, the select-window hook ran.

I added a check to ensure that we don't close the ui doc view if its buffer was the last one visible, and that seems to do the right thing now!

yyoncho:
Thank you! :heart: 
And congrats for your first `lsp-ui` PR! :tada:

PR #258, opened 2019-03-14
[lsp-ui-doc] hide info popup during editing and when point moves

Khady:
It is very inconvenient to use lsp-ui-doc with the popup at point because it doesn't disappear until the next call to hover. It means that the point can end up behind the popup or that the popup will hide some code will typing.

This PR solves the issue by relying on two hooks: post-command-hook and before-change-functions. It tries to be smart to avoid flickering of the popup. The popup is hidden only when the point is moved out of the range concerned by the hover.

I guess that this behavior should be made optional, for all the people who like to keep the child frame alive or those who have the doc inline. It is not clear how I should do that.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -711,6 +711,7 @@ HOVER is the returned signature information."
   "Save the buffer during `pre-command-hook' so that
 `post-command-hook' can check if the buffer is different and
 cleanup the hover state.")
+(defvar lsp-ui-doc--line-saved nil)
 
 (defun lsp-ui-doc--cleanup-hover-state ()
   "Unset state used by hover/eldoc."
@@ -723,6 +724,7 @@ cleanup the hover state.")
 Hide the information popup if the point moves out of the range
 concerned by the hover."
   (setq lsp-ui-doc--command-saved-buffer (current-buffer))
+  (setq lsp-ui-doc--line-saved (line-number-at-pos))
   (unless (and lsp--hover-saved-bounds
                (lsp--point-in-bounds-p lsp--hover-saved-bounds))
     (lsp-ui-doc--cleanup-hover-state)))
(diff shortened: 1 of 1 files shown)


Khady:
If this behaviour is sensible, I'd like to implement the same thing in lsp-mode to hide the highlights.

yyoncho:
> If this behaviour is sensible, I'd like to implement the same thing in lsp-mode to hide the highlights.

Highlights in lsp-mode should be reworked to work independently of eldoc/lsp-ui-doc. E. g. the highlights should be removed once you leave  the identifier not after waiting eldoc/lsp-ui-doc to trigger.

Khady:
Yep. That's what I meant. And it is basically the system that I implemented in this PR. So it would be a duplicated in lsp-mode for highlights. And independent of eldoc/lsp-ui.

Khady:
In the response to textDocument/hover sent by the lsp server you are using, is there a range? There is already some code that should prevent the popup from being closed as long as the point stays in the range. 

I can try to implement a fallback when the range is not provided. By asking emacs the symbol at point. But it will not always be perfect.

yyoncho:
Yes, there is range, check https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L2891

Khady:
My question was about the server part actually. That is used in this part of the PR https://github.com/emacs-lsp/lsp-ui/pull/258/files#diff-8d3d19658f7e3909bab9fcb6ed8117cbR710

But thanks to your link, I'm thinking I could change my code and used the same system as in lsp-mode (save the range to a variable on use `lsp--point-in-bounds-p`).

Khady:
I changed the code to rely on `lsp--hover-saved-bounds`. This variable has to be reset when the popup is hidden. Otherwise it is possible to leave a hover range, to move in again and have no eldoc/popup.

Also I decided to hide the popup when the cursor goes on a different line. Otherwise it be hidden behind the popup.

One last thing I am thinking about, this hiding behavior could be activated only when the popup is displayed at point. If you think it's a good idea.

yyoncho:
I tested again, and the doc frame hides and pops again even if I am still on the identifier.

Khady:
Closing as this is no longer possible since 3b91ef6

PR #259, opened 2019-03-15
[lsp-ui-peek] Handle when server returns single location

yyoncho:
See https://github.com/emacs-lsp/lsp-ui/commit/df65b13630119b4b15f9b4af14cfb578be4d9d05#commitcomment-32763869


sebastiencs:
Thanks @tomterl @yyoncho

PR #263, opened 2019-04-07
Do not enable/disable imenu in lsp-ui-imenu-enable

yyoncho:
Rationale:

- lsp-imenu-enable should be managed by lsp-mode or the user (see https://github.com/emacs-lsp/lsp-mode/issues/729 )
- lsp-ui-imenu works with default handler as well


yyoncho:
I am proceeding with the merge due to the lack of interest to review the PR. The comments will be addressed additionally if any.

kirk86:
Is this `which-func-ff-hook` still and error when using lsp-ui?

PR #269, opened 2019-04-15
Add Terminal Friendly Bindings For Imenu

isomarcte:
The key binding `<return>` only binds to return in the graphical mode of emacs. If emacs is running in the terminal we want to use `RET` instead.


yyoncho:
Thank you, sorry for the delay!

PR #273, opened 2019-04-18
[lsp-ui-doc] Consider no-record flag when hiding the doc frame.

Alexander-Miller:
This change ensures that the doc frame is not hidden when a window is selected only temporarily, e.g. using `with-selected-window`. An example where that happens is treemacs and its follow-mode feature.

An even better solution would have been to use `buffer-list-update-hook`, but I haven't figured out how to get to the current window without introducing some new variable for it.


yyoncho:
Thank you!

PR #277, opened 2019-04-28
Fix typo

danielmartin:
This name changed by means of the typo fix in https://github.com/emacs-lsp/lsp-mode/pull/800.


yyoncho:
we should rename the original method to lsp-cur-line-diagnostics to indicate that the method is used outside of the project.

PR #281, opened 2019-05-02
Bring back lsp-ui-doc-render-function

Gei0r:
`lsp-ui-doc-render-function` was removed in e48c80cf (although the `defvar` was retained).
I need that function to clean up the text encoding.
This PR re-enables calling `lsp-ui-doc-render-function` if it is set.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -228,6 +228,7 @@ Because some variables are buffer local.")
 
 (defun lsp-ui-doc--extract-marked-string (marked-string &optional language)
   "Render the MARKED-STRING."
+  (let ((extracted
   (string-trim-right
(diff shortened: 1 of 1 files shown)


yyoncho:
Isn't it possible to achieve the same via altering "markdown" in lsp-language-id-configuration ?

Gei0r:
Hmm, what I wanted to do was not really doing any real rendering, but actually changing the text to render after the actual rendering.
I had used `lsp-ui-doc-render-function` for that before it was removed.

After more thinking I have come to the conclusion that advising `lsp-ui-doc--extract-marked-string` is the more correct way to do it.

Anyway, the `defvar` for `lsp-ui-doc-render-function` should be removed because it's not used anymore.

PR #290, opened 2019-05-23
lsp-ui-sideline-apply-code-actions: Directly apply in case of just one option

bastianbeischer:
The function used to prompt for the action to apply, even if there was just one
option for the current line.

As far as I can tell there was no other way to just "apply the fixit that you see in the sideline" directly, apart from clicking on it, which I find inconvenient.

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -347,13 +347,7 @@ CURRENT is non-nil when the point is on the symbol."
   (interactive)
   (unless lsp-ui-sideline--code-actions
     (user-error "No code actions on the current line"))
-  (let* ((actions lsp-ui-sideline--code-actions)
-         (title (completing-read "Apply: " (--map (gethash "title" it) actions)
-                                 nil t))
-         (action (--first (equal (gethash "title" it) title) actions)))
-    (unless action
-      (error "Fail to apply action"))
-    (lsp-execute-code-action action)))
+  (lsp-execute-code-action (lsp--select-action lsp-ui-sideline--code-actions)))
 
 (defun lsp-ui-sideline--code-actions (actions bol eol)
   "Show code ACTIONS."


yyoncho:
> As far as I can tell there was no other way to just "apply the fixit that you see in the sideline" directly, apart from clicking on it, which I find inconvenient.

In general, lsp-mode has lsp-execute-code-action which has that behaviour but it does one more roundtrip to the server. Can you utilize lsp--select-action and lsp-execute-action in this PR to unify lsp-mode and lsp-ui?

bastianbeischer:
I can try, but I can't make any promises since I'm not a good lisp programmer :-)

In fact I knew about lsp-execute-code-action and it is actually already used in lsp-ui-sideline-apply-code-actions (with the selected action as argument). But at least to me calling lsp-execute-code-action directly, does not work as well as the lsp-ui-sideline-apply-code-actions variant. Whereas the latter always allows me to apply the code-actions show in the sideline as long as point is in the correct line, the former is a bit particular about the exact position of point. That must be because lsp-code-actions-at-point does not return the wanted actions.

Also, one could argue that unifying lsp-mode with lsp-ui is orthogonal to this PR, which just aims to improve the usability of lsp-ui-sideline-apply-code-actions, in case of only a single action there's no need to prompt the user which one to apply.

yyoncho:
Fair enough, I think that replacing the body with:

```
(lsp-execute-code-action (lsp--select-action lsp-ui-sideline--code-actions))
```

should be ok.

> That must be because lsp-code-actions-at-point does not return the wanted actions.

FTR lsp-ui is always sending the current line as a selection (which IMO is wrong, vscode is not doing that) and a lot of the code actions that the LS is providing are never listed in the sideline. You may fix that by setting `lsp-ui-sideline-update-mode`. This will be fixed in the future.

bastianbeischer:
> Fair enough, I think that replacing the body with:
> 
> ```
> (lsp-execute-code-action (lsp--select-action lsp-ui-sideline--code-actions))
> ```
> 
> should be ok.

That does exactly what I want and looks better, pushed. Thanks!

> 
> > That must be because lsp-code-actions-at-point does not return the wanted actions.
> 
> FTR lsp-ui is always sending the current line as a selection (which IMO is wrong, vscode is not doing that) and a lot of the code actions that the LS is providing are never listed in the sideline. You may fix that by setting `lsp-ui-sideline-update-mode`. This will be fixed in the future.

Thanks for letting me know, I will try setting that variable to 'point'. I agree that lsp-mode (being a library) should be as precise as possible. But on the other hand some fuzziness can be convenient an UI module...

yyoncho:
@bastianbeischer FYI the problem with point is that it wont work fine if you use symbols(it will blick to much). I want to optimize the sideline for that case before making it optional. But this is not going to happen anytime soon.

PR #291, opened 2019-05-30
Allow other flycheck checkers to run

wyuenho:
Many LSP servers such as ts-ls only provide syntax errors and type diagnostics without linting, whereas most flycheck checkers lint. The flycheck checkers and the lsp-ui checkers cannot substitute one another but are complementary instead. 

While it is possible to chain the checkers in individual mode hooks, there's no way to have flycheck to run with chained checkers when the file is first opened and LSP mode is auto-configured, because flycheck is turned on immediately after setting up the checker.

This change gives other checkers a chance to run by default, but also allows the user to turn it off locally or globally if this behavior is not desirable.


wyuenho:
Closing this for now as the next checkers are global instead of per buffer.

PR #302, opened 2019-07-02
Allow the components of lsp-ui to require `lsp-ui.el`.

cireu:


brotzeit:
I've tried this, but the files don't get required this way ?

cireu:
@brotzeit  It should work now.

brotzeit:
Cherry picked. Thanks!

PR #308, opened 2019-07-17
[lsp-ui-doc] add var to kill doc frame each time

mpanarin:
If you are working with several frames already, current implementation produces
nasty issues with numbering of windows. ex. If you have 1 frame with 1 window
and lsp-doc triggered. If you create new frame after that it will have numbering
starting from 3, as number 2 if taken by child frame of ui-doc. Which is very
inconvenient.


mpanarin:
waits for https://github.com/deb0ch/emacs-winum/pull/22
in order to make a proper solution

deb0ch:
Merged upstream ! 🎉

Issue #248, opened 2019-02-28
max-width is not being enforced if there is a link in the doc


ashrithr:
Is there a way to enforce the max-width even when there are hyperlinks in the doc?

I am currently using the following settings:

```
(setq lsp-ui-doc-max-width 40
        lsp-ui-doc-max-height 30)
```

Here is a screenshot:

<img width="816" alt="screenshot 2019-02-27 19 25 10" src="https://user-images.githubusercontent.com/1916107/53539744-ab65f700-3ac7-11e9-9ba0-03abb0414075.png">

PS: It also doesn't seem to take max-height into effect

PR #311, opened 2019-07-31
Fix position of wide frames

kurnevsky:
It calculates wrong position if the buffer is still too wide after resizing.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -68,6 +68,12 @@
                  (const :tag "At point" at-point))
   :group 'lsp-ui-doc)
 
+(defcustom lsp-ui-doc-alignment 'frame
+  "How to align the doc."
+  :type '(choice (const :tag "Frame" frame)
+                 (const :tag "Window" window))
+  :group 'lsp-ui-doc)
+
 (defcustom lsp-ui-doc-border "white"
   "Border color of the frame."
   :type 'color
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you, can you share how to test this PR?

kurnevsky:
I noticed it on my project where I had a very long type written without spaces so it couldn't be wrapped on resizing phase. So you need to display a long string inside this buffer somehow.

yyoncho:
Sorry, but I am not very familiar with that piece of code - is this fix for https://github.com/emacs-lsp/lsp-ui/issues/248 or it addresses only the placement?

kurnevsky:
Here is the original issue:
screenshot-2019-07-31-20:18:46
How it looks after the fix:
screenshot-2019-07-31-20:26:18

Also I took courage to fix it for case when there is another buffer at the right:
screenshot-2019-07-31-20:26:33
After the fix:
screenshot-2019-07-31-20:26:51

yyoncho:
I think that the second part of the change (moving the popup in the current window bounds instead of using current frame bounds) should be configurable because I guess that there are people preferring that location (e. g. when doing a vertical split space in the current window might not be sufficient). @sebastiencs what do you think?

kurnevsky:
> when doing a vertical split space in the current window might not be sufficient

In this case it will be drawn over right window from the beginning of the frame. But ok, ping me if you want it configurable.

kurnevsky:
Made it configurable.

PR #314, opened 2019-08-22
Fix hover not rendering simple MarkedString as Markdown

mpanarin:
According to LSP, when contents is a simple string, it is
still a MarkedString and should be rendered as Markdown
https://microsoft.github.io/language-server-protocol/specification#textDocument_hover

in ex. elixir-ls sends a simple string in contents as its docs are always markdown.


yyoncho:
That you!

PR #324, opened 2019-10-15
Fix ability to turn off doc comments

JohnC32:
If doc comments, the doc shown at the top of the buffer, is present and you turn off doc comments
via (lsp-ui-doc-enable nil), the comments are now removed.


yyoncho:
Thank you!

JohnC32:
Great. Thanks for the quick update!

Issue #297, opened 2019-06-18
Left fringe in lsp-ui-doc frame


seagle0128:
- GNU Emacs 27.0.50
- macOS Mojave
- lsp-ui 20190523 

I noticed there is a left fringe in `lsp-ui-doc` child frame, but no right fringe. I checked `lsp-ui-doc-frame-parameters` and am not sure if it's as designed. The fringe also brings incorrect color render after changing the color theme, e.g from dark to light theme, the left fringe is still dark.

yyoncho:
Fixed with #325

PR #325, opened 2019-10-16
Fix fringe color in lsp-ui-doc frame

### LINKED ISSUE #151: [lsp-ui-doc]: Strange left border on child frame
I've been seeing a strange left border on my child frame that looks like a title bar, but is empty otherwise. 

screen shot 2018-06-22 at 16 01 12

Any idea where this is coming from?

Thanks!

Schnouki:
Themes that set a background color for the fringe look weird as that color is also used for the fringe inside the lsp-ui-doc frame. Just setting it to `nil` in the frame is enough to go back to something normal.

Fixes #151, #157 and maybe #297.

---
Examples with the Zenburn theme:

- without this patch:
    
- with this patch:


yyoncho:
Thank you for fixing this!

PR #328, opened 2019-10-23
Do not hover document after lsp-ui-doc-hide

### LINKED ISSUE #327: Not to hover document after lsp-ui-doc-hide
When binding some key to `lsp-ui-doc-hide` and pressing the key, document is hovered by `post-command-hook`.

I think it makes sense not to do so.

iquiw:
so that child frame will not be shown just after hidden.

fixes #327


yyoncho:
Thank you!

PR #333, opened 2019-11-07
Add a customization option to control whether to show the directory in peek

r-darwish:


danielmartin:
The customization makes sense and the code looks good. I think this is ready to land.

PR #336, opened 2019-11-15
Add a new command lsp-ui-doc-glance

liuchong:
Take a glance on doc, trigger display hover information popup and hide it on next typing.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -753,5 +753,16 @@ It is supposed to be called from `lsp-ui--toggle'"
   (interactive)
   (lsp-ui-doc--hide-frame))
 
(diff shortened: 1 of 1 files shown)


liuchong:
ping

lsp-ui-doc.el
```diff
@@ -753,5 +753,16 @@ It is supposed to be called from `lsp-ui--toggle'"
   (interactive)
   (lsp-ui-doc--hide-frame))
 
+(defun lsp-ui-doc--glance-hide-frame ()
```
brotzeit: Shouldn't this function also be interactive ?

danielmartin:
Thanks for the contribution!

PR #342, opened 2019-11-20
Should filter CR before displaying it to either sidebar or doc

kiennq:
Some language server are returning the document including CR characters, we should filter it out since it's annoying and does nothing even if presented.
Part of code refactor from emacs-lsp/lsp-mode#1190


yyoncho:
Thank you!

PR #348, opened 2019-12-06
Split long lines in diagnostics

chessman:
It also removes unnecessary blank lines between messages. `offset` was incremented twice in `lsp-ui-sideline--find-line`.

Fixes: #150

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -186,7 +186,7 @@ BOL & EOL are beginning and ending of the user point line.
 if UP is non-nil, it loops on the previous lines.
 if OFFSET is non-nil, it starts search OFFSET lines from user point line."
   (let ((win-width (lsp-ui-sideline--window-width))
-        (index (if (null offset) 1 (1+ offset)))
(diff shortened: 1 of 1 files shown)


lsp-ui-sideline.el
```diff
@@ -318,14 +318,28 @@ CURRENT is non-nil when the point is on the symbol."
         (when (overlay-get ov 'current)
           (lsp-ui-sideline--toggle-current ov nil))))))
 
+(defun lsp-ui-sideline--split-long-lines (lines)
+  "Show diagnostics on the current line."
```
danielmartin:
```suggestion
  "Fill LINES so that they are not longer than 100 characters."
```

lsp-ui-sideline.el
```diff
@@ -318,14 +318,28 @@ CURRENT is non-nil when the point is on the symbol."
...
+  (mapcan (lambda (line)
+            (let* ((max-length 100))
```
danielmartin: Maybe we could consider making the ` 100` a variable? It depends on the available screen space, so some users may want to tweak it.

PR #351, opened 2019-12-12
lsp-ui-doc: wrap text when rendering inline

aconchillo:
The window-body-width is used instead of lsp-ui-doc-max-width.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -225,18 +225,21 @@ Because some variables are buffer local.")
 
 (defun lsp-ui-doc--inline-wrapped-line (string)
   "Wraps a line of text for inline display."
-  (cond ((string-empty-p string) "")
-        ((< (length string) lsp-ui-doc-max-width) string)
-        (t (concat (substring string 0 (- lsp-ui-doc-max-width 1))
-                   "\\\n"
-                   (lsp-ui-doc--inline-wrapped-line (substring string (- lsp-ui-doc-max-width 1)))))))
+  (let ((doc-max-width (lsp-ui-doc--inline-window-width)))
+    (cond ((string-empty-p string) "")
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -225,18 +225,21 @@ Because some variables are buffer local.")
 
 (defun lsp-ui-doc--inline-wrapped-line (string)
   "Wraps a line of text for inline display."
-  (cond ((string-empty-p string) "")
-        ((< (length string) lsp-ui-doc-max-width) string)
-        (t (concat (substring string 0 (- lsp-ui-doc-max-width 1))
-                   "\\\n"
-                   (lsp-ui-doc--inline-wrapped-line (substring string (- lsp-ui-doc-max-width 1)))))))
+  (let ((doc-max-width (window-body-width)))
```
danielmartin: Could you test this with `(lsp-ui-doc--inline-window-width)` instead of `(window-body-width)`? The former is a little bit more complete and takes into account line number width, etc. See the differences on my system, if I enable line numbers:

`(window-body-width)`:

Screen Shot 2019-12-13 at 1 53 42 AM

(Note that some letters from "annotation" are missing.)

`(lsp-ui-doc--inline-window-width)`:

Screen Shot 2019-12-13 at 1 55 57 AM

aconchillo: Updated! That seems to fix the issue. However wrapping is still not great in some cases, it would be great if the temp-buffer used by lsp to render could use `fill-paragraph` then we wouldn't need this. but I couldn't make it work.

danielmartin: Yes, I also feel like this code is reimplementing `fill-region`, but I don't see it as a blocker to merge the patch. Thanks!

PR #362, opened 2019-12-26
Fix lsp-ui-doc strange backslash rendering

danielmartin:
In some cases, `lsp-ui-doc` showed strange backslash characters after https://github.com/emacs-lsp/lsp-ui/pull/351 was merged.


kurnevsky:
Works fine now, thanks!

PR #363, opened 2019-12-27
Improve the positioning of documentation at point

danielmartin:
Introduce a new algorithm to prefer positioning the documentation child frame on top of the
symbol at point, if there's space, so as to avoid obstructing the code
that follows.

Some screenshots:

**Default behavior:**

Screen Shot 2019-12-27 at 7 28 52 PM

**No space above the symbol:**

Screen Shot 2019-12-27 at 7 31 33 PM

**No horizontal space (child frame is moved to the left to fit the parent frame):**

Screen Shot 2019-12-27 at 7 32 06 PM

**Child frames may be shown between window splits, if necessary** 

Screen Shot 2019-12-27 at 7 34 42 PM

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -61,7 +61,7 @@
   :type 'boolean
   :group 'lsp-ui-doc)
 
-(defcustom lsp-ui-doc-position 'top
+(defcustom lsp-ui-doc-position 'at-point
   "Where to display the doc."
   :type '(choice (const :tag "Top" top)
                  (const :tag "Bottom" bottom)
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -61,15 +61,16 @@
   :type 'boolean
   :group 'lsp-ui-doc)
 
-(defcustom lsp-ui-doc-position 'top
+(defcustom lsp-ui-doc-position 'at-point
```
kurnevsky: Is it intentional change?

danielmartin: Yes, @yyoncho suggested in Gitter that we could make it the default. Let's see the feedback we get as people upgrade the package.

Issue #347, opened 2019-12-04
After fixing an error and save buffer, lsp-ui still displays the error message


hisnawi:
I am enabling flycheck syntax checking only on save with :

>      flycheck-check-syntax-automatically  '(mode-enabled save)

And disabling lsp ui live reporting:

>           (setq lsp-ui-flycheck-live-reporting nil)

however, if I follow the steps:
edit a line and cause an error -> save buffer -> lsp-ui shows error message -> fix error -> save buffer -> lsp-ui still shows same error message 

If I check flycheck list of errors the error is not listed there anymore which tells me it is a lsp-ui issue.

melton1968:
I was able to work around this issue with the following one line patch to `lsp-ui-flycheck.el`:

```
diff --git a/lsp-ui-flycheck.el b/lsp-ui-flycheck.el
index 3c48710..c5d5cb4 100644
--- a/lsp-ui-flycheck.el
+++ b/lsp-ui-flycheck.el
@@ -216,7 +216,6 @@ See https://github.com/emacs-lsp/lsp-mode."

 (defun lsp-ui-flycheck--report nil
   (and flycheck-mode
-       lsp-ui-flycheck-live-reporting
        (flycheck-buffer)))

 ;; FIXME: Provide a way to disable lsp-ui-flycheck
```

I am unfamiliar with the code-base, so I am uncertain if this is the correct fix, but it is working for me with the configuration listed above. Without this change, there is no way that flycheck can keep the errors messages in sync with the last result from the language server.

If you have installed lsp-ui from elpa, you can simply delete the line from the file and remove the corresponding .elc file and then restart emacs.

hisnawi:
Hmm. it dosen't seem that worked for me.
I removed the line and reloaded emacs, but I still see live updates by flycheck.

melton1968:
After a little more testing, I see that in general you are correct.  At this point, this is what I observe:

At global scope, I get live error reporting. 

In inner scopes (e.g. object abc { … }), I get the desired behavior.

So, the fix I suggested is clearly not correct, although it does makes the syntax checking much more usable for me since I am typing in some inner scope much more often that at global scope. I will continue with it until the real fix is in place.

ps. I presume you removed the .elc file?

hisnawi:
There was no .elc file. Did you mean in the elpa/lsp-ui(version-number) directory?

melton1968:
[...]

PR #365, opened 2019-12-29
Issue #347 If `lsp-ui-flycheck-live-reporting` is nil, respect `flycheck-check-syntax-automatically`

melton1968:
Changes in `lsp-ui-flycheck` in order to support three configurations:

1. `lsp-ui-flycheck-live-reporting` is not `nil`:
    update errors after every diagnostic report

2. `lsp-ui-flycheck-live-reporting` is `nil` and `flycheck-check-syntax-automatically` contains `'save`:
    update errors after a diagnostic report is the buffer is unmodified

3. `lsp-ui-flycheck-live-reporting` is `nil` and `flycheck-check-syntax-automatically` does not contain `'save`:
    never update errors

### DIFF
--- a/lsp-ui-flycheck.el
+++ b/lsp-ui-flycheck.el
@@ -225,7 +225,9 @@ See https://github.com/emacs-lsp/lsp-mode."
 ;; FIXME: Provide a way to disable lsp-ui-flycheck
 (defun lsp-ui-flycheck-enable (_)
   "Enable flycheck integration for the current buffer."
-  (setq-local lsp-ui-flycheck--save-mode (memq 'save flycheck-check-syntax-automatically))
+  (setq-local lsp-ui-flycheck--save-mode
+	       (or (memq 'save flycheck-check-syntax-automatically)
+		   lsp-ui-flycheck--save-mode))
   (setq-local flycheck-check-syntax-automatically nil)
   (setq-local flycheck-checker 'lsp-ui)
   (lsp-ui-flycheck-add-mode major-mode)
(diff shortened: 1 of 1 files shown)


lsp-ui-flycheck.el
```diff
@@ -215,15 +216,17 @@ See https://github.com/emacs-lsp/lsp-mode."
...
+	   (and lsp-ui-flycheck--save-mode (not (buffer-modified-p))))
        (flycheck-buffer)))
 
 ;; FIXME: Provide a way to disable lsp-ui-flycheck
 (defun lsp-ui-flycheck-enable (_)
   "Enable flycheck integration for the current buffer."
-  (when lsp-ui-flycheck-live-reporting
-    (setq-local flycheck-check-syntax-automatically nil))
+  (setq-local lsp-ui-flycheck--save-mode (memq 'save flycheck-check-syntax-automatically))
+  (setq-local flycheck-check-syntax-automatically nil)
```
yyoncho: do we need that when live reporting is nil? The effect ATM is that if you call 2 times lsp in the same buffer the on-save action stops to work. If it is mandatory, I suggest changing the previous line to

`(setq-local lsp-ui-flycheck--save-mode (or (memq 'save flycheck-check-syntax-automatically) lsp-ui-flycheck--save-mode))

melton1968:
We do need to set `flycehck-check-syntax-automatically` to nil, otherwise, flycheck will run on save before the diagnostics are ready. 

I made the change you suggested so that invoking lsp on the buffer multiple times is idempotent with respect to the local variables.

cheers,
mark

yyoncho:
Thank you for taking care of this and sorry for the delay: I was off a few days and then forgot about this PR.

Issue #369, opened 2020-01-02
lsp-ui-doc childframe does not respect altered lsp-ui-doc-background


svenssonaxel:
To reproduce:
- Open with `emacs -Q`
- Evaluate:
```
(progn
  (defvar bootstrap-version)
  (let ((bootstrap-file
	 (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
	(bootstrap-version 5))
    (unless (file-exists-p bootstrap-file)
      (with-current-buffer
          (url-retrieve-synchronously
           "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
           'silent 'inhibit-cookies)
	(goto-char (point-max))
	(eval-print-last-sexp)))
    (load bootstrap-file nil 'nomessage))
  (straight-use-package 'lsp-mode)
  (straight-use-package 'lsp-ui)
  (require 'lsp-mode)
  (require 'lsp-ui)
  (setq 
   lsp-ui-doc-enable t
   lsp-ui-doc-header t
   lsp-ui-doc-include-signature t
   lsp-ui-doc-position 'bottom
   lsp-ui-doc-alignment 'window
   lsp-ui-doc-use-childframe t
   lsp-ui-doc-use-webkit nil
   lsp-ui-doc-delay 0.2

   lsp-ui-sideline-enable nil
   lsp-ui-imenu-enable t)
  (custom-set-faces
   '(lsp-ui-doc-background ((t :background "#ff00ff")))
   '(lsp-ui-doc-header ((t :foreground "#ff8000" :background "#00ff00")))
   '(lsp-ui-doc-url ((t :inherit link))))
  )
```
- Open a new file ~/test.js
- `M-x lsp`
- Write `document.activeelement` and wait for the info-box.
- It is magenta, as expected.
- Evaluate:
```
  (custom-set-faces
   '(lsp-ui-doc-background ((t :background "#0000ff")))
   '(lsp-ui-doc-header ((t :foreground "#00ff80" :background "#808080")))
   '(lsp-ui-doc-url ((t :inherit link))))
```
- Activate the documentation frame again.
- It seems that the changed `lsp-ui-doc-header` is respected, but not the new value for `lsp-ui-doc-background`.

svenssonaxel:
The following diff is a first attempt to fix it. It works both inline and with childframe. Note that
- It is not tested with webkit.
- In inline mode, it will allow other features than :background to be used if set.

```
[...]

PR #371, opened 2020-01-05
WIP: Fix for #369

svenssonaxel:
Work in progress: Fix for issue #369.

Theme changes won't update background color and fringe width
automatically. Previously, there was advice to delete frame after
load-theme. This commit changes this to instead update background
color and fringe width after load-theme and also after
custom-set-faces.

Note especially that
- This fix is not tested with webkit.
- If other features than :background are set in lsp-ui-doc-background
  face, then the inline mode might use them. Added documentation to
  advice against this.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -117,7 +117,7 @@ option."
   '((((background light)) :background "#b3b3b3")
     (t :background "#272A36"))
   "Background color of the documentation.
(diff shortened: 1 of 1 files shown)


brotzeit:
I will try to test this the next days.

tap349:
This problem still exists: changing color for `lsp-ui-doc-background` face has no effect on childframe

Issue #322, opened 2019-10-14
lsp-ui very slow typing with Vue vls (emacs 27 with native JSON)


garyo:
I'm trying to use lsp and lsp-ui with Vue's vls. I have emacs 27 nightly with libjansson support (confirmed that's working OK).
Typing in a Vue file is extremely slow; maybe one character echoes per second -- sometimes even slower. 
Here's my config:
```
       (use-package lsp-mode
         :commands lsp
         :hook ((vue-mode . lsp)
                (typescript-mode . lsp)
                (python-mode . lsp))
         :config
         (setq lsp-prefer-flymake nil
               lsp-log-io nil
               lsp-trace nil
               lsp-print-performance t
               lsp-response-timeout 20
               )
         )
       (use-package lsp-ui
         :commands lsp-ui-mode
         :hook (lsp-mode . lsp-ui-mode)
         :config
         (setq lsp-ui-doc-enable t
               lsp-ui-doc-use-childframe t
               lsp-ui-doc-position 'top
               lsp-ui-doc-include-signature t
               lsp-ui-sideline-enable nil
               lsp-ui-flycheck-enable t
               lsp-ui-flycheck-list-position 'right
               lsp-ui-flycheck-live-reporting t
               lsp-ui-peek-enable t
               lsp-ui-peek-list-width 60
               lsp-ui-peek-peek-height 25
               )
	 )
       (use-package company-lsp
         :config
         (push 'company-lsp company-backends)
         ;;XXX only for javascript-typescript (add to mode hook)
         ;(make-local-variable 'company-transformers)
         ;; (add-to-list 'company-transformers 'lsp-prefix-company-transformer)
         ;; Disable client-side cache because the LSP server does a better job.
         (setq company-lsp-async t
               company-lsp-cache-candidates 'auto
               ;; this is important for typescript language server which returns way too much!
               ;; also javascript-typescript-langserver, same thing
               company-lsp-match-candidate-predicate #'company-lsp-match-candidate-flex
               )
         )
       (use-package helm-lsp
         )
```
and here's a profiler run while I was just trying to type 10 chars:
```
	- lsp-ui-doc--make-request                                         33  70%
	 - lsp--send-request-async                                         33  70%
	  - lsp--flush-delayed-changes                                     33  70%
	   - -each                                                         33  70%
	    - #<compiled 0x2b84fa1>                                        33  70%
	     - lsp-notify                                                  33  70%
	      - lsp--send-notification                                     33  70%
[...]

PR #375, opened 2020-01-10
lsp-ui-doc: Delay request instead of delaying display

PythonNut:
Since making and sending requests is a nontrivial operation, running it via `post-command-hook` can cause significant slowdowns when typing. Since `lsp-ui-doc` frames are already delayed, I've reordered things such that the request is created and sent after the idle delay instead of before it. 

If many commands are called in a short period, most of the idle timers will be cancelled. Previously, lots of requests would have been made, but their docs would ultimately be ignored. Now those requests won't be made in the first place.

I'm not sure if this is related to #322, but I was personally experiencing a lot of lag while typing with the `texlab` server, with this change, the lag is significantly reduced.


brotzeit:
Ok let's give this a try.

PR #381, opened 2020-01-29
Allow to focus into lsp-ui-doc-frame with mouse and keyboard

### LINKED ISSUE #376: lsp-ui-doc scroll popup functions
Hi people!

I've seen the discussion on #155 about scrolling the popup but I am addin a new issue here to track it better.

I currently can show/hide the popup but I still NEED the mouse to scroll and that is very non-emacsy. I don't know elisp/emacs frameworks too much to help on this but this would be very nice.

Thanks in advance for looking into it :)

kiennq:
This should fix #376 
With this fix:
- We can focus into `lsp-ui-doc-frame` with mouse and thus, can scroll it
- We can bind key to focus into `lsp-ui-doc-frame` , and unfocus from it with keyboard.

Here is how it looks like
8fa081dc-29df-4ad4-a44f-45e2e709897b

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -34,6 +34,7 @@
 (require 'dash-functional)
 (require 'goto-addr)
 (require 'markdown-mode)
+(require 'cl-lib)
 
 (when (featurep 'xwidget-internal)
   (require 'xwidget))
@@ -133,7 +134,6 @@ Only the `background' is used in this face."
 
 (defvar lsp-ui-doc-frame-parameters
   '((left . -1)
-    (no-accept-focus . t)
     (no-focus-on-map . t)
     (min-width  . 0)
     (width  . 0)
@@ -152,7 +152,6 @@ Only the `background' is used in this face."
     (visibility . nil)
     (mouse-wheel-frame . nil)
     (no-other-frame . t)
-    (cursor-type . nil)
     (inhibit-double-buffering . t)
     (drag-internal-border . t)
     (no-special-glyphs . t)
@@ -634,6 +633,8 @@ HEIGHT is the documentation number of lines."
                   buffer
                   `((child-frame-parameters . ,params))))
          (frame (window-frame window)))
+    (with-current-buffer buffer
+      (lsp-ui-doc-frame-mode 1))
     (set-frame-parameter nil 'lsp-ui-doc-buffer buffer)
     (set-window-dedicated-p window t)
     (redirect-frame-focus frame (frame-parent frame))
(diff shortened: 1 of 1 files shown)


kiennq:
Ping, can someone help review this?

lsp-ui-doc.el
```diff
@@ -773,14 +773,40 @@ It is supposed to be called from `lsp-ui--toggle'"
 
 (defun lsp-ui-doc--glance-hide-frame ()
   "Hook to hide hover information popup for lsp-ui-doc-glance."
-  (lsp-ui-doc-hide)
-  (remove-hook 'pre-command-hook 'lsp-ui-doc--glance-hide-frame))
+  (when (or (overlayp lsp-ui-doc--inline-ov)
+            (let ((frame (lsp-ui-doc--get-frame)))
+              (and frame (frame-visible-p frame)
+                   (not (frame-focus-state frame)))))
```
brotzeit: Maybe make this a lambda ? This is a little hard to read.

kiennq: Can you give me an example of how should I change this?

brotzeit: I would put the and condition in a lambda similiar to
```
(let ((func (lambda (frame)
           (and frame (frame-visible-p frame)
                (not (frame-focus-state frame)))))
          (frame (lsp-ui-doc--get-frame)))
(funcall func frame)
(when ...)
```
and put the lambda function where now the let begins. 
Sorry I just woke up, but I think you get what I mean =)

yyoncho: there is `cl-letf` which is better than let + funcall.

kiennq: Updated!

kiennq: Updated!

brotzeit: I don't really see the difference in this case. I thought `cl-letf` is used to temporarily override a function defintion.

yyoncho: Its my mistake - I was over the phone - the correct macro is cl-flet.

kiennq: Both of them work, however using cl-letf allow additional normal let binding in the same command, so it would prefer it over `cl-flet`. Also since `cl-flet` doesn't do recursive binding, `cl-labels` may be a better choice for that case.

yyoncho:
Merging as it is. The `cl-flet` vs `cl-flet` vs `cl-labels` discussion should not block this important PR.

kurnevsky:
Not sure if it's related to this PR but lsp-ui frame now gets focus when mouse cursor is over frame when it pops up, which is not very convenient. I don't remember it was happening before, but will test more a bit later, if needed.

PR #386, opened 2020-02-05
Add clang-tidy specific error explainer for the lsp-ui checker

danielmartin:
New feature for C++ language servers that emit `clang-tidy` messages (like Clangd).

Press `e` on a `clang-tidy` error and you'll see an explanation from the LLVM website, so that you can learn about the error message before applying a fix-it (if available).

**Screenshot:**

Screen Shot 2020-02-05 at 10 35 46 PM

### DIFF
--- a/lsp-ui-flycheck-clang-tidy.el
+++ b/lsp-ui-flycheck-clang-tidy.el
@@ -1,0 +1,120 @@
+;;; lsp-ui-flycheck-clang-tidy.el --- clang-tidy specific Flycheck support for lsp-mode -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 Daniel Martín
+;; URL: https://github.com/emacs-lsp/lsp-ui
+;; Keywords: lsp, ui
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


lsp-ui-flycheck-clang-tidy.el
```diff
@@ -0,0 +1,120 @@
...
+                       (format
+                        "Error accessing clang-tidy documentation: %s"
+                        (error-message-string error-status)))
+                    (let ((doc-contents
+                           (lsp-ui-flycheck-clang-tidy--extract-relevant-doc-section)))
+                      (lsp-ui-flycheck-clang-tidy--explain-error
+                       #'shr-insert-document doc-contents)))))
+  "Loading documentation...")
+
+(defun lsp-ui-flycheck-clang-tidy-error-explainer (error)
```
yyoncho: autoloaded?

lsp-ui-flycheck.el
```diff
@@ -208,7 +209,10 @@ See https://github.com/emacs-lsp/lsp-mode."
   :start #'lsp-ui-flycheck--start
   :modes '(python-mode) ; Need a default mode
   :predicate (lambda () lsp-mode)
-  :error-explainer (lambda (e) (flycheck-error-message e)))
+  :error-explainer (lambda (e)
```
yyoncho: We have to figure out where to put this logic - ATM I am moving the flycheck integration in lsp-mode. Is it possible to alter the flycheck checker once it is defined?

danielmartin: I think so, but I'd need to check the Flycheck API. Where do you want to move the flycheck integration?

yyoncho: What do you think about having `lsp-clangd.el` in lsp-mode to host this functionality? Later we could use the same code for `lsp-treemacs-errors-list`.

yyoncho: we may keep it as it is and use `functionp` to access this function if it is present. Of course, I am open to alternative proposals.

danielmartin: I'll open a PR against the `lsp-mode` repo, now that the logic has moved.

danielmartin:
Implemented in lsp-mode repo: https://github.com/emacs-lsp/lsp-mode/pull/1406

PR #388, opened 2020-02-08
Fix double-spacing in multiline diagnostics

raxod502:
Before:

image

After:

image

I don't really understand why the code was the way it was to begin with, so it's entirely possible that this change is wrong and breaks some other use case. I would welcome suggestions.


brotzeit:
Hmmm, looks reasonale. I'm trying to add some tests.

brotzeit:
I'm merging this now. I will work in this the next das.

PR #392, opened 2020-02-10
Remove usage of frame-focus-state

### LINKED ISSUE #391: void-function frame-focus-state in Emacs 26.3
I'm on Emacs 26.3. Presumably after #381 I'm now getting

```
Error in post-command-hook (lsp-ui-doc--glance-hide-frame): (void-function frame-focus-state)
```

when the doc popup tries to go away. frame-focus-state doesn't appear to exist in frame.el for Emacs 26.

kiennq:
This fixes #391 
We actually don't need to check the `frame-focus-state` since if the child frame is being focused, `lsp-ui-doc--get-frame` will return `nil` due to there's no child frame's child frame.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -207,10 +207,15 @@ Because some variables are buffer local.")
   "Set the frame parameter ‘lsp-ui-doc-frame’ to FRAME."
   `(set-frame-parameter nil 'lsp-ui-doc-frame ,frame))
 
-(defun lsp-ui-doc--get-frame (&optional include-deleted-frame)
+(defun lsp-ui-doc--get-frame (&optional _)
   "Return the child frame."
   (let ((frame (frame-parameter nil 'lsp-ui-doc-frame)))
     (and (frame-live-p frame) frame)))
+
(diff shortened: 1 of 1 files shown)


brotzeit:
I'm having troubles to write working tests for the lsp-ui popups,sidlines etc. Does anybody know if this actually works the same way when emacs is in noninteractive mode ?

yyoncho:
@brotzeit I had some troubles in the past when I was writing tests for file watches - I doubt that we will be able to run integration tests that do screen calculation on travis. I hoped that https://github.com/emacs-lsp/lsp-mode/issues/1081 might help but I haven't investigated it. IMO adding byte-compilation tests similar to the one we have in lsp-mode will help a lot - they catch a lot of issues.

danielmartin:
What kind of issues? If `lsp-ui` things don't work well when Emacs is run in batch mode, maybe we can replace the affected Emacs functions in tests using `cl-letf` to return fake values. I used a similar approach for one of my packages.

yyoncho:
> What kind of issues?

Issues like this, calling a function with a wrong number of arguments, undefined function, missing require package, etc.

Using cl-letf will work.

PR #399, opened 2020-02-13
Make lsp-ui-doc-glance works again after #375

kiennq:
- Using `lsp-ui-doc--make-request` in `lsp-ui-doc-glance` instead of `lsp-ui-doc-show`.
With that, even with change in #375, the `lsp-ui-doc--callback` will not be called before command `lsp-ui-doc-glance` finish, thus make the `lsp-ui-doc--glance-hide-frame` (hook in `post-command-hook`) not running and hide the popup immediately.
- Put a delay to ensure unfocus from `lsp-ui-doc-frame` when the child frame is hidden. This is done since when using mouse to focus on doc frame, sometimes it takes time for the focus to really change, and the `post-command-hook` maybe kicked in before that.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -777,19 +777,23 @@ It is supposed to be called from `lsp-ui--toggle'"
   (interactive)
   (lsp-ui-doc--hide-frame))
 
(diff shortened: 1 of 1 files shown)


brotzeit:
Nice! I tried it manually this time #400

Issue #217, opened 2019-01-01
lsp-ui-peek-find-workspace-symbol error


r-darwish:
I tried using the above function in two Rust projects. What I get is:

`insert-file-contents-literally: Read error: Is a directory, /home/roeyd/src/alma/src`

`lsp-ui-find-workspace-symbol` works as expected.

[Here](https://github.com/drrlvn/.emacs.d) is my configuration.

ret2libc:
Same error here. Did you find a solution @r-darwish ?

r-darwish:
Nope. The error still occurs

ret2libc:
@MaskRay are you aware of any problem around this? I'm using ccls + lsp-ui in emacs. Is there anything I can do to help debug the issue? Finding symbols in a workspace is really useful and I would really like to see this fixed.

yyoncho:
proposed fix #406

MaskRay:
@r-darwish Do you know why `insert-file-contents-literally` may run on a directory?

@ret2libc I mainly use `lsp-ui-find-workspace-symbol` instead of `lsp-ui-peek-find-workspace-symbol`, because `workspace/symbol` results can be very long. `lsp-ui-peek` may be a bit clumsy when there are numerous results.

PR #406, opened 2020-02-20
Check if :file is really a file

yyoncho:
- fixes https://github.com/emacs-lsp/lsp-ui/issues/217

#

### DIFF
--- a/lsp-ui-peek.el
+++ b/lsp-ui-peek.el
@@ -698,12 +698,18 @@ references.  The function returns a list of `ls-xref-item'."
 Returns item(s)."
   (-when-let* ((locs (lsp-request method params))
                (locs (if (listp locs) locs (if (vectorp locs) (append locs nil) (list locs)))))
-    (mapcar #'lsp-ui-peek--get-xrefs-list
-            (if (gethash "uri" (car locs))
-                ;; Location[]
-                (--group-by (lsp--uri-to-path (gethash "uri" it)) locs)
-              ;; LocationLink[]
-              (--group-by (lsp--uri-to-path (gethash "targetUri" it)) locs)))))
+    (-filter
+     (-lambda ((&plist :file))
+       (or (f-file? file)
+           (ignore
+            (lsp-log "The following file %s is missing, ignoring from the results."
+                     file))))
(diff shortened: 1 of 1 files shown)


MatthewZMD:
A general question. Is there any guidelines or rule of thumb for lsp team on PR code reviews and testing?

yyoncho:
Do you have something particular in mind?

MatthewZMD:
I'm wondering what's the standard way to test the `needs testing` tagged PRs in the lsp-ui context.

brotzeit:
I just added this label to signal that I don't just ignore a pull request ;)

The idea was that somebody needs to test it manually. Automatic tests are always nice, but I think most stuff needs to be tested manually in case of lsp-ui. But maybe we find at least some solutions for writing tests. No tests at all sucks...

MatthewZMD:
Gotcha

lsp-ui-peek.el
```diff
@@ -698,12 +698,18 @@ references.  The function returns a list of `ls-xref-item'."
...
-    (mapcar #'lsp-ui-peek--get-xrefs-list
-            (if (gethash "uri" (car locs))
-                ;; Location[]
-                (--group-by (lsp--uri-to-path (gethash "uri" it)) locs)
-              ;; LocationLink[]
-              (--group-by (lsp--uri-to-path (gethash "targetUri" it)) locs)))))
+    (-filter
+     (-lambda ((&plist :file))
+       (or (f-file? file)
+           (prog1 nil
```
MaskRay: `(prog1 nil` -> `(ignore`

yyoncho: I wonder which one is the recommended solution here: `ignore` is shorter/more in a functional style and slower while prog1 is more imperative/faster and longer. I am interested to hear what elisp experts recommend.

brotzeit: Why is prog1 faster ? No elisp expert at all :P

yyoncho: prog1 is primitive written in C and there is no function invocation.

yyoncho:
Addressed the comment.

PR #407, opened 2020-02-20
Show sideline result when buffer switched

kurnevsky:
Added `:mode 'alive` to hover request so that lsp-ui will show sideline results even if the buffer is not active.

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -441,15 +441,15 @@ from the language server."
           (lsp-ui-sideline--diagnostics bol eol))
         (when (and lsp-ui-sideline-show-code-actions (or (lsp--capability "codeActionProvider")
                                                          (lsp--registered-capability "textDocument/codeAction")))
-          (lsp--send-request-async (lsp--make-request
-                                    "textDocument/codeAction"
-                                    (if (equal lsp-ui-sideline-update-mode 'line)
-                                        (list :textDocument doc-id
-                                              :range (lsp--region-to-range bol eol)
-                                              :context (list :diagnostics (lsp-cur-line-diagnostics)))
-                                      (lsp--text-document-code-action-params)))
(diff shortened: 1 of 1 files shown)


brotzeit:
> even if the buffer is not active.

I'm just looking at `lsp--create-async-callback` and it seems to me as it is the opposite ? But probably I'm missing something here.

```elisp
('alive (lambda (result)
                  (push (cons lsp--cur-workspace result) results)
                  (if (and (eq (length results) count)
                           (buffer-live-p buf))
                      (with-current-buffer buf
                        (handle-result))
                    (lsp-log "Buffer is not alive ignoring response. Method %s." method))))
```

kurnevsky:
It's not, we make requests only if the buffer is active but show them when already made requests and got a response from the server.

On February 23, 2020 1:56:59 PM GMT+02:00, brotzeit <notifications@github.com> wrote:
>> even if the buffer is not active.
>
>I'm just looking at `lsp--create-async-callback` and it seems to me as
>it is the opposite ? But probably I'm missing something here.
>
>```elisp
>('alive (lambda (result)
>                  (push (cons lsp--cur-workspace result) results)
>                  (if (and (eq (length results) count)
>                           (buffer-live-p buf))
>                      (with-current-buffer buf
>                        (handle-result))
>(lsp-log "Buffer is not alive ignoring response. Method %s." method))))
>```
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/emacs-lsp/lsp-ui/pull/407#issuecomment-590061036

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brotzeit:
Sounds like it would possible to add a test for this. I'm currently looking for stuff that can be covered by tests. If you have any ideas...

PR #410, opened 2020-02-21
Fix current buffer in callback

raxod502:
Otherwise if the user changes which buffer is current before the callback is triggered, the callback will act on the wrong buffer, which may cause an error if for example the new buffer does not have a backing file.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -677,11 +677,11 @@ HEIGHT is the documentation number of lines."
                  lsp-ui-doc-delay nil
                  (let ((buf (current-buffer)))
                    (lambda nil
-                     (with-current-buffer buf
+                     (when (equal buf (current-buffer))
                        (lsp--send-request-async
                         (lsp--make-request "textDocument/hover" (lsp--text-document-position-params))
                         (lambda (hover)
-                          (with-current-buffer buf
+                          (when (equal buf (current-buffer))
                             (lsp-ui-doc--callback hover bounds (current-buffer)))))))))))
       (lsp-ui-doc--hide-frame))))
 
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -675,10 +675,14 @@ HEIGHT is the documentation number of lines."
           (setq lsp-ui-doc--timer
                 (run-with-idle-timer
                  lsp-ui-doc-delay nil
-                 (lambda nil
-                   (lsp--send-request-async
-                    (lsp--make-request "textDocument/hover" (lsp--text-document-position-params))
-                    (lambda (hover) (lsp-ui-doc--callback hover bounds (current-buffer))))))))
+                 (let ((buf (current-buffer)))
+                   (lambda nil
+                     (with-current-buffer buf
```
yyoncho: I would propose checking if the new (current-buffer) is the same as buf. If it has changed - discard the change.

kurnevsky: Yes, the same as in https://github.com/emacs-lsp/lsp-ui/blob/e8200e3b72ecb203a854224eaf73a2194cb4ba46/lsp-ui-sideline.el#L518

raxod502: Yep, that's clearly the right thing to do. Thanks for pointing it out!

yyoncho:
Thank you for contributing to `lsp-ui`!

For future reference, it is good to ping after force-push because there is no notification for that(or at least I do not receive such).

raxod502:
Thanks for the pointer, I'll do that next time. I really appreciate the work you all do on lsp-mode. It does an amazing job of turning Emacs into a modern editor that can compete with the likes of VSCode :)

PR #413, opened 2020-02-22
add test for lsp-ui-flycheck-list--update

brotzeit:
I started with this function, because it's an easy start. @yyoncho you moved some stuff to lsp-mode recently, is `lsp-ui-flycheck-list--update` actually still relevant ?

I also included rustic in the tests because I already added a test with a running lsp server to it. I guess I will switch to rust-mode since it has less dependencies. But this way I could copy and paste almost everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,55 @@
+language: emacs-lisp
+dist: trusty
+sudo: required
+
+notifications:
+  email: false
+
+cache:
+  - directories:
+      - $HOME/emacs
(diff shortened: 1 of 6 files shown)


yyoncho:
> @yyoncho you moved some stuff to lsp-mode recently, is lsp-ui-flycheck-list--update actually still relevant ?

yes, `lsp-ui-flycheck-list` is still in lsp-ui, I moved only the actual flycheck-checker.

PR #417, opened 2020-03-02
modify lsp-ui-doc--callback to return t if a doc was found else nil

rileyrg:
I hope this is ok. I've never done a pull request before.

I made a simple change to lsp-ui-doc--callback to return t if a doc was found. This is so I can fallback to another reference if lsp server doesn't find a doc.


MatthewZMD:
Please resolve these conflicts

rileyrg:
I'm new to this stuff so please bear with me. I merged master back into my
rgr-tweaks branch.  is this now ok? It's confusing the hell out of me;)  It
seems to be "ready to merge".

rileyrg:
withdrawing as bigger issues with glance and timer.

PR #419, opened 2020-03-07
Add cancel tokens for lsp-ui requests

yyoncho:
- the fact that we are not sending cancel tokens is causing troubles to some of
the language servers(e. g. clangd).


yyoncho:
Reference: https://www.reddit.com/r/emacs/comments/fc7bhq/lspmodeclangd_extremely_slow_characterbycharacter/

PR #439, opened 2020-04-25
Add github actions CI

ericdallo:
* Migrate from Travis to Github Actions
* Add byte-compile before tests

Related: https://github.com/emacs-lsp/lsp-ui/pull/392#issuecomment-584260494


ericdallo:
GH Actions working: https://github.com/ericdallo/lsp-ui/actions/runs/87933360

brotzeit:
Sorry for the late response. I didn't work with github actions before, but I would say if tests run we can merge it.

lsp-ui.el
```diff
@@ -46,10 +46,10 @@
 (require 'lsp-ui-doc)
 (require 'dash)
 
-(with-eval-after-load 'flycheck
+(when (featurep'flycheck)
```
brotzeit, line 49: I'm not sure if this works as intended. What if flycheck gets required after lsp-ui ?

lsp-ui-flycheck.el
```diff
@@ -2,7 +2,9 @@
 
 ;; Copyright (C) 2017  fmdkdd
 ;; URL: https://github.com/emacs-lsp/lsp-ui
-;; Keywords: lsp, ui
+;; Keywords: languagues, tools
+;; Package-Requires: ((emacs "26.1") (dash "2.14") (dash-functional "1.2.0") (lsp-mode "6.0"))
```
brotzeit, line 6: Why add the requirements to each file ?

ericdallo: I tried to add [package-lint](https://github.com/purcell/package-lint), but for other problems that was not so easy to fix, I gave up, I'll undo this

lsp-ui.el
ericdallo: Yeah, you are right and I don't know a better way to handle this, but, this is a warning from `package-lint` and I gave up to setup it for other problems, I'll undo it too

ericdallo:
Done @brotzeit !

lsp-ui.el
```diff
@@ -1,12 +1,12 @@
...
 
 ;; Copyright (C) 2017 Tobias Pisani
 ;; Copyright (C) 2018 Sebastien Chapuis, Fangrui Song
 
 ;; Author: Sebastien Chapuis <sebastien@chapu.is>, Fangrui Song <i@maskray.me>
-;; Keywords: lsp
+;; Keywords: languages, tools
 ;; URL: https://github.com/emacs-lsp/lsp-ui
-;; Package-Requires: ((emacs "25.1") (dash "2.14") (dash-functional "1.2.0") (lsp-mode "6.0") (markdown-mode "2.3"))
+;; Package-Requires: ((emacs "26.1") (dash "2.14") (dash-functional "1.2.0") (lsp-mode "6.0") (markdown-mode "2.3"))
```
brotzeit, line 9: Is it necessary to drop emacs25 support ?

ericdallo: when using `package-lint` it warns me that we use some function that I don't record now that is not supported on emacs 25, I'll find out what function was

ericdallo: when I run `package-lint-current-buffer` on `lsp-ui-sideline.el`
image

ericdallo: `lsp-ui.el`
```
430:33: error: You should depend on (emacs "26.1") if you need `frame-outer-width'.
432:37: error: You should depend on (emacs "26.1") if you need `frame-outer-width'.
535:15: error: You should depend on (emacs "26.1") if you need `line-number-display-width'.
657:33: error: You should depend on (emacs "26.1") if you need `frame-parent'.
840:21: error: You should depend on (emacs "26.1") if you need `frame-parent'.
```

ericdallo: What do you think @brotzeit ?

brotzeit: So actually lsp-ui already depends on emacs26 ? Hmmm, IMO we should stick with the dependency of lsp-mode. But it seems too late already =)

I guess most lsp-ui users want fancy stuff and use new emacs versions anyway...
@yyoncho  What do you think ?

yyoncho: AFAIK all functionality has emacs25 fallback(or at least used to have). But IMO since most of it relies on emasc 26, either way, it makes sense to introduce emacs 26+ dependency.

brotzeit:
Let's give it a try.

PR #447, opened 2020-05-24
limit inline doc height

scturtle:


scturtle:
In favor of https://github.com/emacs-lsp/lsp-mode/pull/1740.

PR #456, opened 2020-06-10
Add support for code actions on modeline

ericdallo:
This PR adds a new UI for code actions via modeline.

I didn't change the default UI (sideline), so to enable modeline UI, you need to `(setq lsp-ui-code-actions-type 'modeline)`.

* Single code action:
image
* Multiple code actions:
image
* Fallback to unicode if `all-the-icons` is not installed:
image

Currently, all the core logic of code-actions is on `lsp-sideline.el`, that's not good as `sideline` is a UI implementation detail, we may need to refactor that on the future externalizing to a `lsp-code-actions.el` or something.

Closes https://github.com/emacs-lsp/lsp-mode/issues/1237

### DIFF
--- a/lsp-ui-modeline.el
+++ b/lsp-ui-modeline.el
@@ -1,0 +1,101 @@
+;;; lsp-ui-modeline.el --- lsp ui modeline information -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2020 Eric Dallo
+;;
+;; Author: Eric Dallo <ercdll1337@gmail.com>
+;; URL: https://github.com/emacs-lsp/lsp-ui
+;; Keywords: languages, tools
+;; Version: 6.4
+
+;;; License
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


lsp-ui-modeline.el
```diff
@@ -0,0 +1,101 @@
...
+                                        'face 'lsp-ui-modeline-code-action-face)))))
+    (propertize string
+                'help-echo (concat "Apply code actions (s-l a a)\nmouse-1: "
+                                   (if single-action?
+                                       first-action-string
+                                     "select from multiple code actions"))
+                'mouse-face 'mode-line-highlight
+                'local-map (let ((map (make-sparse-keymap)))
+                             (define-key map [down-mouse-1]
+                               #'lsp-ui-sideline-apply-code-actions)
```
ericdallo, line 87: @yyoncho For some weird reason, the keymap is not working when added to modeline, on any other string this seems to work.
Could you help me with that?

ericdallo:
c/c @matklad

yyoncho:
The idea of emacs-lsp/lsp-mode#1237 (better described in https://github.com/emacs-lsp/lsp-mode/issues/1772) is to make lsp-mode self-contained and provide that feature in lsp-mode, not in lsp-ui.

lsp-ui-modeline.el
yyoncho: check lsp--download-status - you should create map using make-mode-line-mouse-map

ericdallo:
Move support to lsp-mode: https://github.com/emacs-lsp/lsp-mode/pull/1781, we should open another PR on `lsp-ui` overriding the modeline if sideline is enabled.

PR #459, opened 2020-06-16
Use posframe library for lsp-ui-doc

### LINKED ISSUE #331: Consider switching lsp-ui-doc to posframe
IMO we should either switch or copy code from the posframe package: https://github.com/tumashu/posframe . In particular, posframe enforces correctly the size of the windows while lsp-ui-doc implementation does not work fine if there is a very long word. - see https://github.com/emacs-lsp/lsp-ui/issues/309 .

I have used the package to implement popups in dap-mode and I was very satisfied with the behaviour of the package.

(cc @tumashu )

### LINKED ISSUE #334: documentation frame wrong size
Using Golang, but apparently this is the same in other languages. Most often have to jump to an empty line before jumping to a position where I want to show the documentation - otherwise the size of the frame will be too narrow, or too short to show all of the documentation.

Sometimes it's correct width though, and sometimes it's as tall (really tall) as the documentation, and sometimes it's reasonably tall and i can scroll with my wheel to read the long docs, which is fine.

From asking at the official chat, it seems that I'm not the only one with this issue.

Sometimes when the frame has been drawn correctly for a doc string, and when a next shorter doc string is shown, it's like the frame just moves right, so the text is placed correctly, but the frame is just transposed to the new (x,y) upper left corner without changing the size, and thus the right border line of the frame is hidden outside of the right end of the buffer.

Sorixelle:
~~Webkit buffers don't work in posframe windows, so the old manual frame
management is kept around for webkit support.~~

~~`lsp-ui-doc-border` was removed, since the internal border is now used
to pad the frame.~~

~~`lsp-ui-doc-mode` now signals an error if Emacs is older than version 26
or is running in the terminal, as posframe does not work under those
conditions.~~

Closes #331, fixes #334.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -37,6 +37,7 @@
 (require 'goto-addr)
 (require 'markdown-mode)
 (require 'cl-lib)
+(require 'posframe)
 
 (when (featurep 'xwidget-internal)
   (require 'xwidget))
@@ -144,29 +145,14 @@ Only the `background' is used in this face."
   :group 'lsp-ui-doc)
 
 (defvar lsp-ui-doc-frame-parameters
-  '((left . -1)
-    (no-focus-on-map . t)
-    (min-width  . 0)
-    (width  . 0)
-    (min-height  . 0)
-    (height  . 0)
-    (internal-border-width . 1)
+  '((no-focus-on-map . t)
     (vertical-scroll-bars . nil)
     (horizontal-scroll-bars . nil)
-    (right-fringe . 0)
-    (menu-bar-lines . 0)
-    (tool-bar-lines . 0)
-    (line-spacing . 0)
-    (unsplittable . t)
-    (undecorated . t)
-    (top . -1)
-    (visibility . nil)
     (mouse-wheel-frame . nil)
-    (no-other-frame . t)
+    (no-accept-focus . nil)
     (inhibit-double-buffering . t)
-    (drag-internal-border . t)
-    (no-special-glyphs . t)
-    (desktop-dont-save . t))
+    (cursor-type . box)
+    (drag-internal-border . t))
   "Frame parameters used to create the frame.")
 
 (defvar lsp-ui-doc-render-function nil
@@ -365,10 +351,7 @@ We don't extract the string that `lps-line' is already displaying."
   (when (overlayp lsp-ui-doc--inline-ov)
     (delete-overlay lsp-ui-doc--inline-ov))
   (when (lsp-ui-doc--get-frame)
-    (unless lsp-ui-doc-use-webkit
-      (lsp-ui-doc--with-buffer
-       (erase-buffer)))
-    (make-frame-invisible (lsp-ui-doc--get-frame))))
+    (posframe-hide (lsp-ui-doc--make-buffer-name))))
 
 (defun lsp-ui-doc--buffer-width ()
   "Calcul the max width of the buffer."
@@ -402,19 +385,6 @@ We don't extract the string that `lps-line' is already displaying."
         (offset-height (round (aref size 1))))
     (xwidget-resize (lsp-ui-doc--webkit-get-xwidget) offset-width offset-height))
   (lsp-ui-doc--move-frame (lsp-ui-doc--get-frame)))
-
-(defun lsp-ui-doc--resize-buffer ()
-  "If the buffer's width is larger than the current frame, resize it."
-  (if lsp-ui-doc-use-webkit
-      (lsp-ui-doc--webkit-execute-script
-       "[document.querySelector('#lsp-ui-webkit').offsetWidth, document.querySelector('#lsp-ui-webkit').offsetHeight];"
-       'lsp-ui-doc--webkit-resize-callback)
-
-    (let* ((frame-width (frame-width))
-           (fill-column (min lsp-ui-doc-max-width (- frame-width 5))))
-      (when (> (lsp-ui-doc--buffer-width) (min lsp-ui-doc-max-width frame-width))
-        (lsp-ui-doc--with-buffer
-         (fill-region (point-min) (point-max)))))))
 
 (defun lsp-ui-doc--mv-at-point (frame width height start-x start-y)
   "Move FRAME to be where the point is.
@@ -459,12 +429,11 @@ FRAME just below the symbol at point."
     (if (eq lsp-ui-doc-position 'at-point)
         (lsp-ui-doc--mv-at-point frame width height left top)
       (set-frame-position frame
-                          (max (- frame-right width 10 (frame-char-width)) 10)
+                          (max (- frame-right width ) 0)
                           (pcase lsp-ui-doc-position
-                            ('top (+ top 10))
+                            ('top top)
                             ('bottom (- (lsp-ui-doc--line-height 'mode-line)
-                                        height
-                                        10)))))))
+                                        height)))))))
 
 (defun lsp-ui-doc--visit-file (filename)
   "Visit FILENAME in the parent frame."
@@ -495,26 +464,31 @@ FN is the function to call on click."
         (lsp-ui-doc--put-click (match-beginning 0) (match-end 0)
                                'browse-url-at-mouse)))))
 
-(defun lsp-ui-doc--render-buffer (string symbol)
-  "Set the buffer with STRING."
+(defvar lsp-ui-doc--render-string nil
+  "The string to render in the documentation popup.")
+(defvar lsp-ui-doc--render-symbol nil
+  "The symbol to render documentation for.")
+
+(defun lsp-ui-doc--render-buffer ()
+  "Set the buffer with `lsp-ui-doc--render-string'."
   (lsp-ui-doc--with-buffer
    (if lsp-ui-doc-use-webkit
        (progn
          (lsp-ui-doc--webkit-execute-script
           (format
            "renderMarkdown('%s', '%s');"
-           symbol
-           (url-hexify-string string))
+           lsp-ui-doc--render-symbol
+           (url-hexify-string lsp-ui-doc--render-string))
           'lsp-ui-doc--webkit-resize-callback))
      (erase-buffer)
      (let ((inline-p (lsp-ui-doc--inline-p)))
        (insert (concat (unless inline-p (propertize "\n" 'face '(:height 0.2)))
-                       (s-trim string)
+                       (s-trim lsp-ui-doc--render-string)
                        (unless inline-p (propertize "\n\n" 'face '(:height 0.3))))))
      (lsp-ui-doc--make-clickable-link))
    (setq-local face-remapping-alist `((header-line lsp-ui-doc-header)))
    (setq-local window-min-height 1)
-   (setq header-line-format (when lsp-ui-doc-header (concat " " symbol))
+   (setq header-line-format (when lsp-ui-doc-header (concat " " lsp-ui-doc--render-symbol))
          mode-line-format nil
          cursor-type nil)))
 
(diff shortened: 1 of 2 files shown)


yyoncho:
> Webkit buffers don't work in posframe windows, so the old manual frame
management is kept around for webkit support.

Can you elaborate on that?

Sorixelle:
The posframe doesn't get resized correctly with the webkit xwidget in it - it just stays as a one character wide and tall block. I haven't yet been able to determine why this is. I'd like to make this work properly before we merge, however.

Sorixelle:
Just pushed some changes that uses posframe to manage webkit buffers. After a bit of discussion in the Gitter room, I've also made posframe a required dependency. Interested to hear the general community's thoughts in this.

lsp-ui-doc.el
```diff
@@ -776,11 +752,15 @@ before, or if the new window is the minibuffer."
...
+    (add-hook 'delete-frame-functions 'lsp-ui-doc--on-delete nil t)
+    (remove-hook 'focus-in-hook 'posframe--redirect-posframe-focus t)
+    (setq posframe-mouse-banish nil))
    (t
     (lsp-ui-doc-hide)
     (remove-hook 'post-command-hook 'lsp-ui-doc--make-request t)
-    (remove-hook 'delete-frame-functions 'lsp-ui-doc--on-delete t))))
+    (remove-hook 'delete-frame-functions 'lsp-ui-doc--on-delete t)
+    (add-hook 'focus-in-hook 'posframe--redirect-posframe-focus nil t)
+    (setq posframe-mouse-banish t))))
```
yyoncho, line 763: I don't like setting it back to t since it may have been nil initially.IMO we should coordinate with upstream what is the desired behaviour.

lsp-ui-doc.el
```diff
@@ -823,7 +803,10 @@ It is supposed to be called from `lsp-ui--toggle'"
   :init-value nil
   :lighter ""
   :group lsp-ui-doc
-  :keymap `(([?q] . lsp-ui-doc-unfocus-frame)))
+  :keymap `(([?q] . lsp-ui-doc-unfocus-frame))
+  (if lsp-ui-doc-frame-mode
```
yyoncho, line 807: if->when no need from progn.

Cask
```diff
@@ -6,6 +6,7 @@
 (depends-on "flycheck")
 (depends-on "lsp-mode")
 (depends-on "markdown-mode")
+(depends-on "posframe")
```
yyoncho, line 9: this should be listed in the package header in lsp-ui.el and removed from here.

lsp-ui-doc.el
Sorixelle: According to tumashu/posframe#54, that option could probably be safely removed from posframe. It seems, according to [documentation](https://github.com/tumashu/posframe#customizing-mouse-pointer-control), that it's intended to be set from init.el, so it can probably be removed from here.

Sorixelle:
Pushed a set of changes resolving the review and the point placement.

lsp-ui-doc.el
```diff
@@ -86,11 +87,6 @@ This only takes effect when `lsp-ui-doc-position' is 'top or 'bottom."
                  (const :tag "Window" window))
   :group 'lsp-ui-doc)
 
-(defcustom lsp-ui-doc-border "white"
-  "Border color of the frame."
-  :type 'color
-  :group 'lsp-ui-doc)
```
kiennq, line 92: Can we keep this option?
Sometime the doc frame is not visible enough, especially when display near caret, so having doc border is helping much more for visibility.
You can set it default to nil though.

Sorixelle: I had to use the border to pad out the frame, since there were issues with the bottom of the frame having 0 padding. If you have any suggestions to pad out the bottom without the border, I'm all ears.

lsp-ui-doc.el
```diff
@@ -776,11 +761,13 @@ before, or if the new window is the minibuffer."
         (cl-callf copy-tree frameset-filter-alist)
         (push '(lsp-ui-doc-frame . :never) frameset-filter-alist)))
     (add-hook 'post-command-hook 'lsp-ui-doc--make-request nil t)
-    (add-hook 'delete-frame-functions 'lsp-ui-doc--on-delete nil t))
+    (add-hook 'delete-frame-functions 'lsp-ui-doc--on-delete nil t)
+    (remove-hook 'focus-in-hook 'posframe--redirect-posframe-focus t))
```
kiennq, line 765: The doc frame focus works really spotty with this PR, especially when using together with `flycheck-mode`.
I have to press `lsp-ui-doc-focus-frame` key bind several times before I can sucessfully focus into doc frame. Moreover, each time it failed, the doc frame will go away and I need to trigger it showing again.

The reason seems to be `focus-in-hook` become local variable and thus `posframe--redirect-posframe-focus` is only removed from local value, and still being triggered in global value (my guess).

I think instead of removing the function from hook, it might be better to just advice reroute it when `lsp-ui-doc-frame-mode` is on. Some thing like this
```suggestion
    (advice-add 'posframe--redirect-posframe-focus
                :before-until (lambda (&rest _)
                                lsp-ui-doc-frame-mode)
                '((name . lsp--posframe-dont-redirect-on-doc-frame)))
```

lsp-ui-doc.el
```diff
@@ -776,11 +761,13 @@ before, or if the new window is the minibuffer."
...
    (t
     (lsp-ui-doc-hide)
     (remove-hook 'post-command-hook 'lsp-ui-doc--make-request t)
-    (remove-hook 'delete-frame-functions 'lsp-ui-doc--on-delete t))))
+    (remove-hook 'delete-frame-functions 'lsp-ui-doc--on-delete t)
+    (add-hook 'focus-in-hook 'posframe--redirect-posframe-focus nil t))))
```
kiennq, line 770:
```suggestion
    (advice-remove 'posframe--redirect-posframe-focus 'lsp--posframe-dont-redirect-on-doc-frame)
```

lsp-ui-doc.el
```diff
@@ -823,7 +810,9 @@ It is supposed to be called from `lsp-ui--toggle'"
   :init-value nil
   :lighter ""
   :group lsp-ui-doc
-  :keymap `(([?q] . lsp-ui-doc-unfocus-frame)))
+  :keymap `(([?q] . lsp-ui-doc-unfocus-frame))
+  (when lsp-ui-doc-frame-mode
```
kiennq, line 814: If you're using my previous suggestion of advising `posframe--redirect-posfram-focus`, this block does needed

Sorixelle:
@kiennq I pushed a commit that uses advice instead of adding/removing hooks. I didn't run into the issue you described in my testing, so let me know if this fixes the issue.

lsp-ui-doc.el
kiennq: Is the bottom without padding a problem?

Here is what it looks like without border
image
and with border.
image

I would argue that the second one is nicer and more distinguish between popup content and buffer content.

I agree without left/right padding, the popup will look ugly. But that can be solved with `left-fringe`, `right-fringe`. For the top padding, it should be character height already. The bottom padding, is that really necessary?

Sorixelle: Unfortunately, that's not the result - it ends up looking like this, where the bottom border is directly under the text with no padding at all.
lsp-ui-doc popup with no bottom padding

kiennq: I mean, choose between having bottom padding but no border (and hard to distinguish between popup content and buffer content), and no bottom padding but there can be border, the later seems better?

You can take a look again at my screenshot, the one doesn't have border is way to hard to see the content.
Even with the bottom padding, the bottom text line can be displayed with just top-half of the text, and without any border, which makes it looks even weirder.

My point is we should not taken away customization, it will break user configurations. If you still want to have padding for all edges, then probably only do it when `lsp-ui-doc-border` is set to `nil`?
It will archive the same look with when there's padding (and no border), but can still retain the old look of having border, but without bottom padding

Sorixelle: Yeah, I'd agree with that. I'll push a change to bring the border colour options back. It's probably worth going over to posframe and seeing if they can add some way to specify top and bottom padding, since that'd probably be a useful feature to have there anyway.

brotzeit:
I've tested it. Seems to work well.

Lenbok:
Please don't abandon support for terminal users.

Sorixelle:
> Please don't abandon support for terminal users.

From memory, terminal Emacs doesn't support creating child frames at all, so this PR isn't going to break anything that wasn't already not working in the terminal.

Lenbok:
@Sorixelle Using lsp-ui from current master in a terminal I currently use lsp-ui to show doc for functions. I just checked out this PR and instead I get the message about requiring graphical mode, so this PR would definitely be a regression for me.

lsp-ui-doc.el
```diff
@@ -488,26 +457,31 @@ FN is the function to call on click."
         (lsp-ui-doc--put-click (match-beginning 0) (match-end 0)
                                'browse-url-at-mouse)))))
 
-(defun lsp-ui-doc--render-buffer (string symbol)
-  "Set the buffer with STRING."
+(defvar lsp-ui-doc--render-string nil
+  "The string to render in the documentation popup.")
+(defvar lsp-ui-doc--render-symbol nil
```
kiennq, line 462: Why you need this two global variables? The original version already pass `symbol` and `string`, why need to change function signature?

Sorixelle: I needed to call `lsp-ui-doc--render-buffer` from the `xwidget-event` callback lambda to solve the problem where webkit frames would be empty the first time they popped up.
https://github.com/emacs-lsp/lsp-ui/blob/61651bac3d3f932b11d549f5a0d557549369aa17/lsp-ui-doc.el#L651-L658

I can't just pass the parameters down to it, because it's only called when the webkit widget is first created, and since it's a lamba, it'll just capture the first values it sees, and won't update on subsequent hovers.

lsp-ui-doc.el
```diff
@@ -763,6 +756,8 @@ before, or if the new window is the minibuffer."
   "Minor mode for showing hover information in child frame."
   :init-value nil
   :group lsp-ui-doc
+  (unless (posframe-workable-p)
```
Lenbok, line 759: Aha - lsp-ui doc display works in my terminal as I have `lsp-ui-doc-use-childframe` set to `nil`, and that tells lsp-ui to use an inline rendering method. 

You need to allow for that possibility in this check.

Sorixelle: Ah, good call - that check would block that one. I'll fix that up.

Lenbok: I think ivy-posframe does it's check each time it's invoked, so if you run emacs daemon with both GUI and TTY clients, it uses posframe in the GUI and falls back to whatever its other method is when invoked from TTY. Maybe you could do that here also, rather than a single up-front check?

Sorixelle: Just pushed a change that should do that.

lsp-ui-doc.el
kiennq: > I needed to call `lsp-ui-doc--render-buffer` from the `xwidget-event` callback lambda to solve the problem where webkit frames would be empty the first time they popped up

Interesting. I wonder how just call `lsp-ui-doc--render-buffer` will resolve that problem?
From what I can see, the global vars are updated in `lsp-ui-doc--display`, which in turn will also call `lsp-ui-doc--render-buffer`.

yyoncho:
Are there more open issues preventing merging this PR?

@Sorixelle can you check the CI? Seems like it is using something that is already deleted.

Sorixelle:
Nothing that I can think of at the moment. Not sure what's going on with the CI - it's pointing to the wrong line number, and failing on a warning. I just moved the minor mode declaration for `lsp-ui-doc-frame-mode` above `lsp-ui-doc-mode` and that shut it up, so hopefully the CI stops complaining too.

lsp-ui-doc.el
kiennq: @Sorixelle

Sorixelle: The thing is, `lsp-ui-doc--render-buffer` inserts the text directly into the webkit widget via a javascript call, so it needs to be called after both the webkit widget has been inserted, and after `lsp-ui-doc.html` has finished loading. That last one in particular isn't guaranteed by the time `lsp-ui-doc--display` has finished executing, but that `'load-changed` event is fired specifically when the page has finished loading, so it's the best thing for us to use, unless there's something else I'm missing.

brotzeit:
I'll merge this and take a look at the test the next days(the tests are now failing for a different reason). @Sorixelle Thanks!

alanz:
FYI I have reverted this in my local usage, it causes the hover window to capture focus, making editing impossible.

Sorixelle:
This isn't an issue I ran into in my testing. If you've got time, open an issue with some extra detail so we can look into it.

seagle0128:
Any benefit from this integration? I don't see any value here, and it introduces many bugs 😭

lsp-ui-doc.el
```diff
@@ -144,29 +145,14 @@ Only the `background' is used in this face."
...
-    (menu-bar-lines . 0)
-    (tool-bar-lines . 0)
-    (line-spacing . 0)
-    (unsplittable . t)
-    (undecorated . t)
-    (top . -1)
-    (visibility . nil)
     (mouse-wheel-frame . nil)
-    (no-other-frame . t)
+    (no-accept-focus . nil)
```
kurnevsky, line 152: Why was it added explicitly?

kiennq: In our original child frame implementation, we removed the `no-accept-focus` so we can focus into child frame to scroll the document. There is little no issue with that approach though, i.e: mouse is not suddenly set focus to other frame.
However, `posframe` add `no-accept-focus` to `t` explicitly, so we need to add back `no-accept-focus . nil` here

kiennq:
Just asking, but what's the value of your `focus-follow-mouse`?
Can you try to it to `nil`?

seagle0128:
Please revert this PR until it's ready.

kurnevsky:
> Just asking, but what's the value of your focus-follow-mouse?
> Can you try to it to nil?

If it's a question to me then I set it to `t` explicitly - I like this behavior and it fits my WM config. That's why I also add `(no-accept-focus . t)` to `lsp-ui-doc-frame-parameters` myself - I don't usually need focus there though even with this setting somehow it can take focus if I click there with mouse.

Issue #464, opened 2020-06-26
Fix posframe usage for lsp-ui-doc


brotzeit:
> it causes the hover window to capture focus, making editing impossible.

> The posframe integration on lsp-ui is slow, due to new frame is being created every time.

Sorixelle:
In my testing, I didn't run into either of these issues while working on #459. If anyone is experiencing these issues, I'd like to get some more details on the environment this is occuring in so I can try and reproduce these.

dmakarov:
This happens to me on macOS 10.15.5: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109)) of 2019-09-02. I don't have any customizations for lsp-ui or posframe group.

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/103494/85916046-1e62b780-b84d-11ea-9d16-c780fc94786e.gif)

Here I move pointer to a method name, lsp-ui-doc pops up and it has focus. I press `q` to switch focus back to the original frame.

yyoncho:
@alanz are you also using macos?

Sorixelle:
Did some testing, and managed to get to the bottom of the focus issues. It seems to be an issue with Cocoa builds of Emacs. This issue doesn't exist in the [emacs-mac](https://github.com/railwaycat/homebrew-emacsmacport) port, so the solution would be to use this. It seems like this is an Emacs bug, and completely out of our control. More info in [https://github.com/tumashu/posframe/issues/30](this) issue in posframe.

yyoncho:
> It seems like this is an Emacs bug, and completely out of our control

It was working in the previous implementation, right?

Sorixelle:
Just reverted the commit, it seems like it did work on Cocoa builds. I'm still unconvinced this is a problem on our side however, as I don't see any reason our code would cause the new frame to get focus.

dmakarov:
This may be a macOS cocoa issue, but the previous non-posframe implementation did work.

Sorixelle:
[...]

PR #465, opened 2020-06-28
posframe-integration: dont create posframe on every request

kiennq:
Per this https://github.com/emacs-lsp/lsp-ui/issues/464#issuecomment-650664735

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -607,8 +607,7 @@ HEIGHT is the documentation number of lines."
     (lsp-ui-doc--render-buffer)
     (if (lsp-ui-doc--inline-p)
         (lsp-ui-doc--inline)
-      (when (or (not lsp-ui-doc-use-webkit)
-                (not (lsp-ui-doc--get-frame)))
+      (unless lsp-ui-doc-use-webkit
         (lsp-ui-doc--set-frame (lsp-ui-doc--make-frame)))
       (unless (frame-visible-p (lsp-ui-doc--get-frame))
         (make-frame-visible (lsp-ui-doc--get-frame))))))
@@ -624,7 +623,6 @@ The structure of INFO is defined in the documentation of `posframe-show'."
 
 (defun lsp-ui-doc--make-frame ()
   "Create the child frame and return it."
-  (lsp-ui-doc--delete-frame)
   (let* ((before-make-frame-hook nil)
         (buffer-name (lsp-ui-doc--make-buffer-name))
         (buffer (get-buffer-create buffer-name))


yyoncho:
cc @Sorixelle

Sorixelle:
Gave this PR a test on my setup, it looks like the windows aren't appearing in the correct place; almost a similar issue to what we were trying to solve with posframe in the first place.

Example

kiennq:
What's your configuration for `lsp-ui-doc`?
The change is simple, and from the logic it just stop calling `lsp-ui-doc--delete-frame` so the `posframe-show` works as expected and doesn't create new frame so often.
If there's problem, I doubt it would be problem on `posframe` itself (and previous child frame implementation), so just switch to `posframe` will  not make the problem going away.

Sorixelle:
Actually, I think this is a problem on posframe. It seems like, from https://github.com/tumashu/posframe/blob/master/posframe.el#L703-L711, that the posframe's position is only updated when the position is different from last time (we're always passing the same poshandler so it doesn't change), and when the parent frame's size changes (which is unlikely to happen).

kiennq:
@Sorixelle The issue is indeed inside `posframe` implementation, specifically with `posframe--set-frame-position`.
I guess you're setting the `lsp-ui-doc-position` to `bottom`, which will trigger the `poshandler` of `posframe-poshandler-frame/window-bottom-right-corner`.
`posframe-show` will call `posframe--set-frame-position` and since the position is the same (relative from right edge and bottom), it will not trigger `set-frame-position` again.
Well, the child frame size does changed, so without calling `set-frame-position`, the position will be wrong.

See this PR https://github.com/tumashu/posframe/pull/65

kiennq:
@sorixelle, I've created PR to fix problem on `posframe` repo, can you test it with your setup?
Note that problem is not happening with default `lsp-ui` setup that show popup at point

Sorixelle:
Positioning is better, but there also seems to be an issue where the window in the frame isn't correctly sized all the time. In this recording, once I focus the frame, I hold down `l` (I use evil) to move through the line. Moving the point right doesn't jump to the next line once it gets to the end, so it seems like the content of the buffer is all one line; just the window isn't fit to the frame and is too small to display it all.

Example

kiennq:
@Sorixelle What's your setup again? Can you have minimal repro for this?
Did this issue not repro without this change?
I doubt it may related to some of your special minor modes that automatically breaks line visually but doesn't inform Emacs so the `fit-frame-to-buffer` does not work.
Actually it seems that the size of frame is calculated based on all of your 3 lines as a single line.

PR #466, opened 2020-06-29
Fix check for content in lsp-ui-sideline--format-info.

kurnevsky:
This fixes sideline for metals.

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -234,7 +234,7 @@ function signature)."
 If the string has a language, we fontify it with the function provided
 by `lsp-mode'.
 MARKED-STRING is the string returned by `lsp-ui-sideline--extract-info'."
-  (when (and marked-string (lsp-marked-string? marked-string))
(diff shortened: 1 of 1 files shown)


lsp-ui-sideline.el
```diff
@@ -234,7 +234,7 @@ function signature)."
 If the string has a language, we fontify it with the function provided
 by `lsp-mode'.
 MARKED-STRING is the string returned by `lsp-ui-sideline--extract-info'."
-  (when (and marked-string (lsp-marked-string? marked-string))
+  (when (and marked-string (or (lsp-marked-string? marked-string) (lsp-markup-content? marked-string)))
```
danielmartin, line 237: Could you update the PR description explaining the problem with the Metals language server (that should be part of the commit description as well)? That'd help us understand some changes in the future if we need to revisit why something was done.

kurnevsky: Updated the commit message.

kurnevsky: Looks like github completely missed my commit so pushed once again.

PR #479, opened 2020-08-01
fix issue in find-prev(next)-reference (regression of d92cf83)

shunlir:
This is a fix for the regression of this change (https://github.com/emacs-lsp/lsp-ui/commit/d92cf83d95c9ca177b735500ead88cf68dc2e2db)
zero-based line number should be used, the LSP textDocument/references returns zero-based line numbers per the LSP spec.
the result of  `line-number-at-pos` is one-based, while the previous used `lsp--cur-line` is zero-based.

### DIFF
--- a/lsp-ui.el
+++ b/lsp-ui.el
@@ -135,7 +135,7 @@ Both should have the form (FILENAME LINE COLUMN)."
 (defun lsp-ui-find-next-reference (&optional extra)
   "Find next reference of the symbol at point."
   (interactive)
-  (let* ((cur (list buffer-file-name (line-number-at-pos) (- (point) (line-beginning-position))))
+  (let* ((cur (list buffer-file-name (- (line-number-at-pos) 1) (- (point) (line-beginning-position))))
          (refs (lsp-ui--reference-triples extra))
          (idx -1)
          (res (-first (lambda (ref) (cl-incf idx) (lsp-ui--location< cur ref)) refs)))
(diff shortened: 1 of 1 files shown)


lsp-ui.el
```diff
@@ -152,7 +152,7 @@ Both should have the form (FILENAME LINE COLUMN)."
 (defun lsp-ui-find-prev-reference (&optional extra)
   "Find previous reference of the symbol at point."
   (interactive)
-  (let* ((cur (list buffer-file-name (line-number-at-pos) (- (point) (line-beginning-position))))
+  (let* ((cur (list buffer-file-name (- (line-number-at-pos) 1) (- (point) (line-beginning-position))))
```
jcs090218, line 155:
```suggestion
  (let* ((cur (list buffer-file-name (1- (line-number-at-pos)) (- (point) (line-beginning-position))))
```

lsp-ui.el
```diff
@@ -135,7 +135,7 @@ Both should have the form (FILENAME LINE COLUMN)."
 (defun lsp-ui-find-next-reference (&optional extra)
   "Find next reference of the symbol at point."
   (interactive)
-  (let* ((cur (list buffer-file-name (line-number-at-pos) (- (point) (line-beginning-position))))
+  (let* ((cur (list buffer-file-name (- (line-number-at-pos) 1) (- (point) (line-beginning-position))))
```
jcs090218, line 138:
```suggestion
  (let* ((cur (list buffer-file-name (1- (line-number-at-pos)) (- (point) (line-beginning-position))))
```

jcs090218 (commented):
Minor.

jcs090218:
This seems to be already fixed in the master branch! Close this PR now! Thanks!

PR #480, opened 2020-08-01
Add posframe integration to lsp-ui-doc

ericdallo:
Another try to https://github.com/emacs-lsp/lsp-ui/pull/459 which was reverted

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -37,6 +37,7 @@
 (require 'goto-addr)
 (require 'markdown-mode)
 (require 'cl-lib)
+(require 'posframe)
 
 (when (featurep 'xwidget-internal)
   (require 'xwidget))
@@ -144,29 +145,14 @@ Only the `background' is used in this face."
   :group 'lsp-ui-doc)
 
 (defvar lsp-ui-doc-frame-parameters
-  '((left . -1)
-    (no-focus-on-map . t)
-    (min-width  . 0)
-    (width  . 0)
-    (min-height  . 0)
-    (height  . 0)
-    (internal-border-width . 1)
+  '((no-focus-on-map . t)
     (vertical-scroll-bars . nil)
     (horizontal-scroll-bars . nil)
-    (right-fringe . 0)
-    (menu-bar-lines . 0)
-    (tool-bar-lines . 0)
-    (line-spacing . 0)
-    (unsplittable . t)
-    (undecorated . t)
-    (top . -1)
-    (visibility . nil)
     (mouse-wheel-frame . nil)
-    (no-other-frame . t)
+    (no-accept-focus . nil)
     (inhibit-double-buffering . t)
-    (drag-internal-border . t)
-    (no-special-glyphs . t)
-    (desktop-dont-save . t))
+    (cursor-type . box)
+    (drag-internal-border . t))
   "Frame parameters used to create the frame.")
 
 (defvar lsp-ui-doc-render-function nil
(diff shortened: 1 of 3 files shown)


ericdallo:
@Sorixelle created this PR trying to fix some issues, we should merge that when we confirm that the issues reported before are fixed.
It works ok for my NixOS - bspwm window manager.
We should try with `xmonad` users and macos users who reported issues with mouse hover.

lsp-ui-doc.el
```diff
@@ -755,7 +749,14 @@ before, or if the new window is the minibuffer."
 (advice-add #'select-window :around #'lsp-ui-doc-hide-frame-on-window-change)
 
 (advice-add 'load-theme :before (lambda (&rest _) (lsp-ui-doc--delete-frame)))
-(add-hook 'window-configuration-change-hook #'lsp-ui-doc--hide-frame)
+
+(defun lsp-ui-doc--hide-checking-buffer ()
```
kiennq, line 753: Why we only hide frame when the current buffer is different from buffer where posframe is invoked?

Another issue is that `lsp-ui-doc--original-buffer-with-frame` is local variable, thus if it's ever set, it will only contain the value of its buffer, so the second check of `(not (eq (current-buffer) lsp-ui-doc--original-buffer-with-frame))` is likely always failed.

ericdallo: You are right I forgot undo the `defvar-local` that I was testing.

> Why we only hide frame when the current buffer is different from buffer where posframe is invoked?

Because if we hide every time `window-configuration-change-hook` is triggered, the frame hides when the window size change, e.g when mini-buffer size change:
lsp-ui-doc-posframe-bug1

kiennq: Is it necessary to hide when `window-configuration-change-hook`?
`company-posframe` doesn't do it neither

yyoncho: > Is it necessary to hide when `window-configuration-change-hook`?
> `company-posframe` doesn't do it neither

comapany-posframe is managed by `company` itself so there is no need company-posframe to handle that.

kiennq: So is it necessary to hide when `window-configuration-change-hook` when in other buffer?
From what I can see, before may be we worried about changing window size will make the mini frame not fully displayable, especially when not in `at-point` mode. Thus it would make sense to hide the frame.

However, with this change, if the current buffer is still in focus, even with window configuration run it will not hide the frame. Given that window configuration can be easily changed due to mini buffer, I think that we don't need to hide doc frame in this hook anymore.

yyoncho: > From what I can see, before may be we worried about changing window size will make the mini frame not fully displayable, especially when not in at-point mode. Thus it would make sense to hide the frame.

I hope that this is not going to happen or at least it won't be a huge concern. IMO the primary reason for hiding the popup should be buffer change and/or no hover info at the point. Even if we decide to handle that case it should not be unconditional like we currently do, but it should be like "1. subscribe for window-configuration-change. 2. Hide lsp-ui-doc if there is no enough space to show the popup" . The problem with current implementation is that it hides the popup on pretty much every frame resizing(e. g. dap-controls-mode, lv/signature, multiline modeline).

kiennq: I agree.
Here is the commit to add the hook https://github.com/emacs-lsp/lsp-ui/commit/2214908d7bc3a128749d31f75ad53f061b469760.

I've tried it without `window-configuration-change-hook` and found no problem. Maybe we don't need that hook at all.

brotzeit:
Merge ?

yyoncho:
I think that all that complained in #465 should test that.

(@seagle0128 was one of them)

ericdallo:
Yes, I tested it on NixOS (bspwm window manager), but it would be better to test on macOS and gnome that had issues.

yyoncho:
It works fine on gnome2.

seagle0128:
I tested on macOS. It's not working well. The screenshot is below.

image

The original is
image

kiennq:
Can this one be merged yet?

ericdallo:
We need to fix @seagle0128 bug or ask for more people to test it, I think

PR #484, opened 2020-08-09
Fix lsp-ui-find-{next,prev}-reference after line numbers became 0 based

MaskRay:


MaskRay:
This has been working well for me.

Issue #478, opened 2020-07-29
lsp-ui-doc popup for function documentation doesn't show up for elixir-ls but works for other language servers


x-ji:
- OS: MacOS 10.15.6
- Emacs: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.7.0, Carbon Version 158 AppKit 1671.6) of 2020-06-14
- lsp-mode: lsp-mode-20200728.1517
- lsp-ui: lsp-ui-20200718.1035

I am at a loss at the behavior that I'm seeing with my Emacs (Spacemacs). I have `lsp-ui-doc-mode` enabled and the popup shows up correctly upon hovering for language servers such as `vls` (Vue.js). However, with `elixir-ls`, nothing shows up.

I set `lsp-log-io` to `t` and looked at the logs. I saw that `textDocument/hover` seems to have been correctly triggered and a response is received, e.g. 

```
[Trace - 09:35:52 PM] Sending request 'textDocument/hover - (190)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jx/test/backend/test/support/data_case.ex"
  },
  "position": {
    "line": 64,
    "character": 5
  }
}


[Trace - 09:35:52 PM] Received response 'textDocument/hover - (190)' in 31ms.
Result: {
  "range": {
    "start": {
      "line": 64,
      "character": 2
    },
    "end": {
      "line": 64,
      "character": 5
    }
  },
[...]

PR #485, opened 2020-08-11
Add support for Strings as MarkedStrings

J3RN:
Much of the context for this change can be found here: #478 

To recap: [The LSP spec for `MarkedString`](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover) is a union type of `string` and an object of the form `{ language: string; value: string }`. [The definition of the `MarkedString` type in `lsp-protocol.el`](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-protocol.el#L470) only supports the object and not `string`. Presently, `lsp-mode` has no support for union types.

In order to support this feature of the LSP spec, I added a check to see if the content received was a `string`, and if so treat it as a valid `MarkedString` containing markdown. This should be the correct behaviour as the LSP spec states:

```
 * MarkedString can be used to render human readable text. It is either a markdown string
 * or a code-block that provides a language and a code snippet.
```

This change also fixes the precise concern (elixir-ls not displaying documentation pop-ups) of #478.

image

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -300,6 +300,10 @@ We don't extract the string that `lps-line' is already displaying."
          (lsp:marked-string-language contents))
     (lsp-ui-doc--extract-marked-string (lsp:marked-string-value contents)
                                        (lsp:marked-string-language contents)))
+   ;; The specification for MarkedString also includes raw strings of
+   ;; markdown, which is not reflected by `lsp-marked-string?'
+   ((stringp contents)
+    (lsp-ui-doc--extract-marked-string contents lsp/markup-kind-markdown))
    ((lsp-marked-string? contents) (lsp-ui-doc--extract-marked-string contents))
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -300,6 +300,10 @@ We don't extract the string that `lps-line' is already displaying."
          (lsp:marked-string-language contents))
     (lsp-ui-doc--extract-marked-string (lsp:marked-string-value contents)
                                        (lsp:marked-string-language contents)))
+   ;; The specification for MarkedString also includes raw strings of
+   ;; markdown, which is not reflected by `lsp-marked-string?'
+   ((eq (type-of contents) 'string)
```
yyoncho, line 305: nit use: `stringp`

J3RN: Would you believe that I searched for this function and couldn't find it? I was looking for `string-p`. I'll make this change.

PR #487, opened 2020-08-15
Allow lsp-ui-doc-max-width to be nil for frame-width

cfraizer:
Allow for `lsp-ui-doc-max-width` value of `nil` to use the `frame-width` value.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -93,7 +93,7 @@ This only takes effect when `lsp-ui-doc-position' is 'top or 'bottom."
   :group 'lsp-ui-doc)
 
 (defcustom lsp-ui-doc-max-width 150
-  "Maximum number of columns of the frame."
+  "Maximum number of columns of the frame or nil for ``frame-width''."
   :type 'integer
   :group 'lsp-ui-doc)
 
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -93,7 +93,7 @@ This only takes effect when `lsp-ui-doc-position' is 'top or 'bottom."
   :group 'lsp-ui-doc)
 
 (defcustom lsp-ui-doc-max-width 150
-  "Maximum number of columns of the frame."
+  "Maximum number of columns of the frame or nil for ``frame-width''."
```
jcs090218, line 96:
```suggestion
  "Maximum number of columns of the frame or nil for `frame-width'."
```

lsp-ui-doc.el
```diff
@@ -93,7 +93,7 @@ This only takes effect when `lsp-ui-doc-position' is 'top or 'bottom."
   :group 'lsp-ui-doc)
 
 (defcustom lsp-ui-doc-max-width 150
-  "Maximum number of columns of the frame."
+  "Maximum number of columns of the frame or nil for `frame-width'."
   :type 'integer
```
jcs090218, line 97: We should specify a type to `nil` too.

PR #490, opened 2020-08-29
[lsp-ui-sideline] Don't re-render hovers/diagnostics on same line

### LINKED ISSUE #434: Don't re-draw hints if the text is the same
**Describe the bug**
Right now, the hints are re-drawn every time the cursor moves. This results in everything constantly flickering as you move around.

Even if performance were perfect, there's no need to replace already drawn text with exactly the same text. Can we do a simple equality check before drawing?

sebastiencs:
Render hover informations and diagnostics only when changing lines.
Only actions are updated when moving point on the same line

Fixes #434

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -543,6 +543,9 @@ from the language server."
            (lsp--text-document-code-action-params))
          (lambda (actions) (lsp-ui-sideline--code-actions actions bol eol))
          :mode 'tick
+         :error-handler
+         (lambda (&rest)
+           (lsp-ui-sideline--delete-kind 'actions))
          :cancel-token :lsp-ui-code-actions))
       ;; Go through all symbols and request hover information.  Note that the symbols are
       ;; traversed backwards as `forward-symbol' with a positive argument will jump just past the
@@ -567,17 +570,23 @@ from the language server."
                 (push (list symbol bounds (list :line (1- line-widen) :character (- (point) bol))) symbols))))
           (if (null symbols)
               (lsp-ui-sideline--delete-kind 'info)
-            (let ((last-symbol (1- (length symbols)))
+            (let ((length-symbols (length symbols))
+                  (current-index 0)
                   list-infos)
-              (--each-indexed symbols
-                (-let (((symbol bounds position) it)
-                       (index it-index))
+              (--each symbols
+                (-let (((symbol bounds position) it))
                   (lsp-request-async
                    "textDocument/hover"
                    (lsp-make-hover-params :text-document doc-id :position position)
                    (lambda (info)
+                     (setq current-index (1+ current-index))
                      (and info (push (list symbol bounds info) list-infos))
-                     (when (= index last-symbol)
+                     (when (= current-index length-symbols)
+                       (lsp-ui-sideline--display-all-info list-infos tag bol eol)))
+                   :error-handler
+                   (lambda (&rest _)
+                     (setq current-index (1+ current-index))
+                     (when (= current-index length-symbols)
                        (lsp-ui-sideline--display-all-info list-infos tag bol eol)))
                    :mode 'tick))))))))))
 
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -77,10 +77,10 @@
   :group 'lsp-ui-sideline)
 
 (defcustom lsp-ui-sideline-update-mode 'point
-  "Define the mode for updating sideline information.
+  "Define the mode for updating sideline actions.
 
-When set to `line' the information will be updated when user
-changes current line otherwise the information will be updated
+When set to `line' the actions will be updated when user
+changes current line otherwise the actions will be updated
 when user changes current point."
   :type '(choice (const line)
                  (const point))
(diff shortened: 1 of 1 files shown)


yyoncho:
> and diagnostics

I haven't tested the PR, but the diagnostics might be updated if the server sends a new set of them. In this case, we trigger the rerender of the sideline which would require rerendering the new diagnostics.

sebastiencs:
Thanks, I didn't take that into account.
> we trigger the rerender of the sideline

What function is called when this happen ? From where ?
Are you referring to this hook:
```elisp
(add-hook 'flycheck-after-syntax-check-hook 'lsp-ui-sideline--diagnostics-changed nil t)
```

Right now `lsp-ui-sideline` is calling flycheck functions to get diagnostics, which make it useless when flymake is used.
Is there a variable from `lsp-mode` to pick the diagnostics ?

yyoncho:
> What function is called when this happen ? From where ?
> Are you referring to this hook:
> 
> ```emacs-lisp
> (add-hook 'flycheck-after-syntax-check-hook 'lsp-ui-sideline--diagnostics-changed nil t)
> ```

Yes.

> Also, right now `lsp-ui-sideline` is calling flycheck functions to get diagnostics, which make it useless when flymake is used.

There is `lsp-after-diagnostics-hook` - it runs if there is a change in diagnostics in any buffer. 

> Is there a variable from `lsp-mode` to pick the diagnostics ?

Check - lsp--get-buffer-diagnostics.

sebastiencs:
@yyoncho  I set a hook:
```elisp
(add-hook 'lsp-diagnostics-updated-hook 'my-function nil t)
```
in `my-function` I'm calling `lsp--get-buffer-diagnostics` but it returns always `nil`.
Calling manually `lsp--get-buffer-diagnostics` in the buffer returns always `nil` too, even if there are errors in the buffer.

Am I missing something ?

yyoncho:
> in `my-function` I'm calling `lsp--get-buffer-diagnostics` but it returns always `nil`.
> Calling manually `lsp--get-buffer-diagnostics` in the buffer returns always `nil` too, even if there are errors in the buffer.
> 
> Am I missing something ?

The only guess that I have is that the errors in the buffer do not come from lsp-mode. We are using this function to feed flycheck with diagnostics.  I tested locally and it returns results when invoked in buffer with diagnostics,

> (add-hook 'lsp-diagnostics-updated-hook 'my-function nil t)

Most likely you should not use local = t. The hook might not be called in the target buffer. We do that because finding the buffer corresponding to a uri is an extremely slow operation.

sebastiencs:
> The only guess that I have is that the errors in the buffer do not come from lsp-mode. We are using this function to feed flycheck with diagnostics. I tested locally and it returns results when invoked in buffer with diagnostics,

Thanks, you were right the diagnostics didn't come from `lsp-mode`.

I will keep using `flycheck`, for this PR, because flycheck handle buffer modifications and move diagnostics automatically. 

I updated the PR, it will avoid any flickering, even when typing.
I also changed the default of `lsp-ui-sideline-diagnostic-max-lines` from `20` to `1`: sideline is supposed to be just a hint, and also for #494 

If you have some time, or someone else, to provide some feedback, please do.

lsp-ui-sideline.el
```diff
@@ -404,114 +418,129 @@ Push sideline overlays on `lsp-ui-sideline--ovs'."
...
+            (--each-indexed symbols
+              (-let (((symbol bounds position) it)
+                     (index it-index))
+                (lsp-request-async
+                 "textDocument/hover"
+                 (lsp-make-hover-params :text-document doc-id :position position)
+                 (lambda (info)
+                   (and info (push (list symbol bounds info) list-infos))
+                   (when (= index last-symbol)
+                     (lsp-ui-sideline--display-all-info list-infos tag bol eol)))
```
sebastiencs, line 542: This code assumes that the server will always respond, even if it's an empty response.
It also assumes that `lsp-request-async` will always call the callback, even on empty response

yyoncho: 1. it won't be called in case of an error. In this case, `:error-handler` will be called.
2. In won't be called if the buffer is edited(this is what `:mode 'tick` is for). In this case `:cancel-handler` will be called.

yyoncho:
> If you have some time, or someone else, to test the branch, please do.

I will be able to test but most likely tomorrow.

sebastiencs:
> I will be able to test but most likely tomorrow.

Thanks, `lsp-ui-sideline` is easy to break/misbehave so some tests/feedbacks before merging would be great

lsp-ui-sideline.el
sebastiencs: Thanks

yyoncho (approved):
I tested it and it works very well.

yyoncho:
I saw the following error:

```
 Error processing message (wrong-number-of-arguments (((symbols ("new" (169 . 172) (:line 10 :character 8)) ("String" (173 . 179) (:line 10 :character 12))) (doc-id :uri "file:///home/yyoncho/Sources/lsp/lsp-mode/test/fixtures/org-mode/java-project/src/test/java/temp/AppTest.java") (line-modified . t) (this-line . "        new String()") (new-tick . t) (line-changed) (this-tick . 1312) (line-widen . 11) (line . 11) (tag 11 . 180) (bol . 161) (eol . 181) (this-line . "        new String()") (eol . 181) (bol . 161) t) (&rest) (lsp-ui-sideline--delete-kind 'actions)) 1).
```

(FYI I am using lsp-mode with plists instead of hts).

sebastiencs:
Great, thanks for testing.

> I saw the following error:

It should be fixed with [`abcd0ed` (#490)](https://github.com/emacs-lsp/lsp-ui/pull/490/commits/abcd0edebf2f91abf1c6430732782e1d748d3cb0)

PR #498, opened 2020-09-05
Add indent level to with-ui-doc.

jcs090218:
I have added the indent level to `lsp-ui-doc--with-buffer` so you can see things better IMO.


sebastiencs:
Thanks @jcs090218

PR #499, opened 2020-09-05
Fixed some docstring.

jcs090218:
I have fixed some of the docstring, and there are also many functions that are undocumented.

There is one function `lsp-ui-doc--sideline-pos-y` is unused, is it for external use case?


MatthewZMD:
Thank you for your contribution!

> There is one function lsp-ui-doc--sideline-pos-y is unused, is it for external use case?

iirc nothing using this function.

PR #501, opened 2020-09-12
Define key map directly.

jcs090218:
Is there a reason that we don't declare key map variable directly? 😕 

Thanks!


sebastiencs:
Thanks ! I don't see any reason

PR #503, opened 2020-09-12
Some logic clean up and better with defense programming.

jcs090218:
This patch does the following changes.

#### For imenu

* Some logic clean up
* Use defense programming strategy to `lsp-ui-imenu--get-padding`.
* Supply missing docstring. (By my understanding)

#### For sideline

* Some operations changed.
* Removed unnecessary newlines that may cause confusions.

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -636,8 +636,7 @@ COMMAND is `company-pseudo-tooltip-frontend' parameter."
 This does not toggle display of flycheck diagnostics or code actions."
   (interactive)
   (when (bound-and-true-p lsp-ui-sideline-mode)
-    (setq lsp-ui-sideline-show-hover
-          (not lsp-ui-sideline-show-hover))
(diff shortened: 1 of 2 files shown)


lsp-ui-sideline.el
```diff
@@ -579,13 +579,13 @@ from the language server."
...
                    (lambda (info)
-                     (setq current-index (1+ current-index))
+                     (cl-incf current-index)
                      (and info (push (list symbol bounds info) list-infos))
                      (when (= current-index length-symbols)
                        (lsp-ui-sideline--display-all-info list-infos tag bol eol)))
                    :error-handler
                    (lambda (&rest _)
-                     (setq current-index (1+ current-index))
+                     (cl-incf current-index)
```
brotzeit, line 588: I think we need to require cl-lib for this right ?

jcs090218: Yeah, we are requiring `lsp-mode`; and the module itself already requires `cl-lib`. :)

jcs090218:
I think this is a simple patch with code improvement. Merge this now. :)

Issue #414, opened 2020-02-25
Child frames grab input events


raxod502:
Using the childframe-based interface for LSP popups, having the mouse cursor over the childframe causes all input events to be directed to the childframe's buffer, usually causing errors because that buffer is read-only and will not react correctly to most user commands. This is extremely frustrating as you can be typing and then suddenly a documentation popup will appear and all further editing commands become no-ops (they try to act on the wrong buffer) until you realize what's going on and move the mouse.

According to the [Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Input-Focus.html), this behavior may vary depending on the window manager in use. I am running Manjaro Linux KDE with the default window manager.

Workaround:

```elisp
(setq lsp-ui-doc-use-childframe nil)
```

kiennq:
May be it's similar to this issue https://github.com/emacs-lsp/lsp-ui/pull/381#issuecomment-585598390?
Will setting `focus-follows-mouse` help?

raxod502:
Setting `focus-follows-mouse` does not help, but this solves the problem:

```elisp
(add-to-list 'lsp-ui-doc-frame-parameters '(no-accept-focus . t))
```

Perchance this should just be added to lsp-ui. It seems semantically correct, and you can even still scroll in the childframe to see the rest of the text (although doing so _will_ focus the frame, which is not super desirable -- although this could be argued both ways).

raxod502:
Actually, I am now observing the problem even with that setting. I must have done something wrong when I tested previously.

samwalls:
Is there any update on this? It makes emacs very annoying to use in an i3 context and you happen to leave your cursor in the window, particularly if the window is small.

I would use the workaround `(setq lsp-ui-doc-use-childframe nil)`. However, for some reason, whenever the documentation pops up, it changes the size of the frame ever so slightly, making all the text jump around. What could be causing this?

kiennq:
@samwalls Did you try what raxod502 mentioned? Setting `lsp-ui-doc-frame-parameters` to include `no-accept-focus`. Or setting `focus-follows-mouse` to `nil`.

kiennq:
Also, see https://github.com/i3/i3/issues/1127

samwalls:
Hey @kiennq, I've just been trying to get that to work. Using spacemacs, it seems like it _does_ work, although it causes an error on startup when I add it to `dotspacemacs/user-config`:
```
[...]

PR #504, opened 2020-09-12
lsp-ui-doc: make cursor shows on keyboard focus

kiennq:
Fix issue mentioned in https://github.com/emacs-lsp/lsp-ui/issues/414#issuecomment-691512719

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -1137,19 +1137,24 @@ It is supposed to be called from `lsp-ui--toggle'"
 (defun lsp-ui-doc-focus-frame ()
   "Focus into lsp-ui-doc-frame."
   (interactive)
-  (when (lsp-ui-doc--frame-visible-p)
-    (set-frame-parameter (lsp-ui-doc--get-frame) 'lsp-ui-doc--no-focus nil)
+  (when-let* ((frame (lsp-ui-doc--get-frame))
+              (visible (lsp-ui-doc--frame-visible-p)))
+    (set-frame-parameter frame 'lsp-ui-doc--no-focus nil)
+    (set-frame-parameter frame 'cursor-type t)
     (lsp-ui-doc--with-buffer
-      (setq cursor-type t))
-    (select-frame-set-input-focus (lsp-ui-doc--get-frame))))
+      (setq cursor-type 'box))
+    (select-frame-set-input-focus frame)))
 
 (defun lsp-ui-doc-unfocus-frame ()
   "Unfocus from lsp-ui-doc-frame."
   (interactive)
   (-some-> (frame-parent) select-frame-set-input-focus)
   (when-let* ((frame (lsp-ui-doc--get-frame)))
+    (set-frame-parameter frame 'lsp-ui-doc--no-focus t)
+    (set-frame-parameter frame 'cursor-type nil)
+    (lsp-ui-doc--with-buffer
+      (setq cursor-type nil))
     (when lsp-ui-doc--from-mouse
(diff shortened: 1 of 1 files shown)


kiennq:
@sebastiencs

sebastiencs:
Thanks @kiennq, I think removing this line fixes your issue: https://github.com/emacs-lsp/lsp-ui/blob/7873160d3ad33081bf6dc25a68430e83133d880e/lsp-ui-doc.el#L804 

I don't remember why I added it and it seems that removing it doesn't cause undesired behavior.
It's a simpler solution that modifying the parameter `cursor-type` as you propose.

Can you confirm ?

kiennq:
Well, remove the frame parameter at that time also fine.
But when you focus the doc frame by mouse, the cursor will show up, having the cursor-type there will help hiding that.
Why using keyboard to navigate doc frame need cursor, using mouse doesn't require that (rather should not display text cursor in that case).

sebastiencs:
You are right, modifying the frame parameter gives us more control. I will merge the PR.
Can you just reset the variable (not frame parameter) `cursor-type` to `nil` in `lsp-ui-doc-unfocus-frame`.

As a side note, it should not be necessary anymore to click with the mouse on the frame to be able to scroll it.
Just moving the mouse pointer over the frame, and using the mouse wheel will scroll the documentation. (Without any click).
Can you observe the same behavior ? It depends on the windows manager and I didn't get any feedback about that change

kiennq:
> Can you just reset the variable (not frame parameter) `cursor-type` to `nil` in `lsp-ui-doc-unfocus-frame`.

While it requires both local variable `cursor-type` and frame parameter `cursor-type` to showing the cursor, just setting frame parameter one to `nil` only will hide it.
That's why the cursor is not showing up in the old code.

I think we can just stop assigning local `cursor-type`, to save a few CPU cycles. But that can be done later.
For now I reset the variable as you suggested.



> As a side note, it should not be necessary anymore to click with the mouse on the frame to be able to scroll it.
> Just moving the mouse pointer over the frame, and using the mouse wheel will scroll the documentation. (Without any click).
> Can you observe the same behavior ? It depends on the windows manager and I didn't get any feedback about that change

No, I didn't observer that behavior. I'm on Windows though.


Another thing is that the `lsp-ui-doc--no-focus` is reset only when `lsp-ui-doc--from-mouse`, is that okay?
Does that affect the case where I focus on doc frame once and unfocus it (by keyboard), then later whenever doc frame show up, mouse will be captured inside it?

sebastiencs:
> I think we can just stop assigning local cursor-type, to save a few CPU cycles. But that can be done later.

Agreed, if setting the frame parameter is enough there is no need to set the variable.

> No, I didn't observer that behavior. I'm on Windows though.

Ah, windows is a bit particular I guess

> Another thing is that the lsp-ui-doc--no-focus is reset only when lsp-ui-doc--from-mouse, is that okay?

mmh I don't think it's intended. It shouldn't depend on `lsp-ui-doc--from-mouse`.
Can you change that as well ?

kiennq:
> Can you change that as well ?

Done

sebastiencs:
@kiennq I just test your branch and I still see the cursor after `lsp-ui-doc-unfocus-frame`, or just by scrolling with the mouse wheel.
Looks like the variable `cursor-type` is still necessary

kiennq:
@sebastiencs Can you check it with latest one? The `cursor-type` is not set to `t` in doc frame anymore

sebastiencs:
Thanks @kiennq, it's working now

PR #506, opened 2020-09-13
lsp-ui-doc: send hover request only for supported servers

leungbk:
----

#

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -1007,6 +1007,7 @@ before, or if the new window is the minibuffer."
            :cancel-token :lsp-ui-doc-hover))))))
 
 (defun lsp-ui-doc--handle-mouse-movement (event)
+  "Show the documentation corresponding to the text under EVENT."
   (interactive "e")
   (when lsp-ui-doc-show-with-mouse
     (and (timerp lsp-ui-doc--timer-mouse-movement)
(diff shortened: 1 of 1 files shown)


sebastiencs:
@leungbk Thanks !

@yyoncho `lsp--capability` does not seem to be a function that return quickly, considering that `lsp-ui-doc--handle-mouse-movement` will be call on _every_ mouse movement.

Can the return value of `(lsp--capability :hoverProvider)` changed during the session ?
Or we can cache it in a variable.

yyoncho:
> Can the return value of `(lsp--capability :hoverProvider)` changed during the session ?
> Or we can cache it in a variable.

It can if you have 2 servers connected to the buffer and kill one of them it might change. Not sure if we should be concerned about this case. 

BTW it is better to use `lsp-feature?` instead of `lsp--capability`. `lsp-feature?` can be optimized a lot as it is.

yyoncho:
And one more case - I think that might lsp-ui will be initialized before the server has returned its capabilities.

sebastiencs:
@yyoncho Right, thanks !

@leungbk Can you move that check to `lsp-ui-doc--mouse-display`.
`lsp-ui-doc--handle-mouse-movement` should remains as fast as possible.

sebastiencs:
Thank you @leungbk

Issue #171, opened 2018-08-24
lsp-ui sideline is displaying symbol info in wrong order


yyoncho:
There are servers which handle the requests asynchronously and they might return the symbol info, not in the order they have been called (e. g. JDT LS). The result is that `lsp-ui` displays the symbol information in the wrong order.

brotzeit:
Is this still an issue ?

yyoncho:
![wrong-order](https://user-images.githubusercontent.com/13259670/76147058-52994900-60a1-11ea-82b2-d459dc23bc1a.png)


Yes. In this case, obtaining `String` data is slowest and is always the last symbol in the list.

yyoncho:
Just want to mention that using emacs-promise/emacs-deferred will be very helpful for such issues: https://github.com/emacs-lsp/lsp-mode/issues/1471

kurnevsky:
If I got it right it can't be shown with strict ordering. We don't know the line length that we will receive from the server and this length affects the line number where it can be displayed.

yyoncho:
> If I got it right it can't be shown with strict ordering. We don't know the line length that we will receive from the server and this length affects the line number where it can be displayed.

Why? We do async calls and we can render the elements in the same order as they are in the current line.

kurnevsky:
Because if we got a too wide response which can't be displayed on the next line (because it won't fit) it will find a line with narrower text where it can be displayed. So before we received a response we can't know the exact line where it'll be shown and the only way to make all lines in order is to shift previously displayed lines.

yyoncho:
Hm, are you saying the wrong order is caused by the length of the item, not on the order of receiving the response?

kurnevsky:
It's async calls that make these lines out of order. Regarding different line lengths - not sure what happens. I'd assume that we just skip those lines where the response can't fit and leave them empty, but need to check/test it. But my point is that we can't simply make them in order just because we can't know in advance which lines we should skip.

yyoncho:
Can we make sure that we render the symbols once we get them all and start rendering/placing from the first one and make sure that the symbols after the first symbol are placed bellow it? It may even reduce blinking/noise.

yyoncho:
Apart from that, we should not refresh the symbols when the user moves the cursor on the same line.

kurnevsky:
> Can we make sure that we render the symbols once we get them all and start rendering/placing from the first one and make sure that the symbols after the first symbol are placed bellow it?

[...]

PR #509, opened 2020-09-14
Add variable to control whether to wait for all symbols

kurnevsky:
As described here https://github.com/emacs-lsp/lsp-ui/issues/171#issuecomment-691627065
cc @sebastiencs

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -111,6 +111,11 @@ when user changes current point."
   :type '(choice file (const :tag "Disable" nil))
   :group 'lsp-ui-sideline)
 
+(defcustom lsp-ui-sideline-wait-for-all-symbols t
+  "Wait for all symbols before displaying info in sideline."
+  :type 'boolean
+  :group 'lsp-ui-sideline)
+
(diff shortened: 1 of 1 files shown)


sebastiencs:
Thanks @kurnevsky.
I don't have a server to test with slow responses, but if you move to a line, wait for the first item to show up, move the cursor on the right or left immediately, are the rest of the items displayed ?

My concern is that in `lsp-ui-sideline--display-all-info` there this condition:
```elisp
  (when (and (eq (current-buffer) buffer)
             (equal tag (lsp-ui-sideline--calculate-tag))
             (not (lsp-ui-sideline--stop-p)))
```

The `tag` variable include `(point)`, so if you move the cursor while waiting for some items to be displayed, they might never be displayed, since this condition will fail.

Can you confirm ?

kurnevsky:
@sebastiencs yes, you are right. But it's happening even with `lsp-ui-sideline-wait-for-all-symbols` set to `t` (or without this change at all) - when I move cursor left or right before I receive all responses sideline isn't shown at all. So probably it should be fixed separately.

sebastiencs:
@kurnevsky Ah, right. I will fix that. Thanks !

PR #513, opened 2020-09-19
Fix lsp-ui-peek-find-definition changing origin buffer start window

ericdallo:
When the same buffer is splitted and I use `lsp-ui-peek-find-definition`, the left window(origin buffer) start is changed incorrectly

Gif:
lsp-ui-peek-bug


sebastiencs:
Thank you @ericdallo

Issue #511, opened 2020-09-16
lsp-ui-imenu auto-refresh


ulises:
`lsp-ui-imenu` is great to quickly see the layout of a Java class however the buffer is not updated once the accompanying buffer changes, i.e. visiting a different class. 

The `imenu` buffer should auto-update when this happens (like its `treemacs` counterpart).

brotzeit:
Duplicate of #295

brotzeit:
I think we should set `imenu-auto-rescan` to t by default for lsp-ui.

jcs090218:
Sounds like a good idea! :)

brotzeit:
But I'm not sure which hook we should use to reload the imenu entries.

jcs090218:
It feel like we have to setup the rule for this... feature?

brotzeit:
> there is one implementation in lsp-treemacs-symbols which uses a timer. IMO it should be reworked to use lsp-on-idle-hook

Maybe we can add a generic solution that can be used for different use cases.

minikN:
Should setting `imenu-auto-rescan`  to `t` manually fix this? Because for me doesn't have any effect. Still not auto refreshing.

jcs090218:
> Should setting imenu-auto-rescan to t manually fix this? Because for me doesn't have any effect. Still not auto refreshing.

Would calling `lsp-ui-imenu` again do the trick? 😕 The current implementation should work in this scenario.

vrx:
yes it does

it is possible to launch it when file is saved in a specific mode (cpp in this case)
note the last argument set to true of add-hook in order to apply it to the local-mode
and select-window to reset keyboard focus where it was originally

```
(add-hook 'c++-mode-hook
            '(lambda ()
               (add-hook 'after-save-hook (lambda ()
                                            (let ((cur-window (selected-window)))
                                              (lsp-ui-imenu)
                                              (select-window cur-window)
                                              )
                                            ) 0 t)))
```

jcs090218:
Got it! Thanks for confirming.

I'm working on #515. I think this issue would be resolved after this PR is merged!

jcs090218:
After commit 982b4669721c8bcd4dcf2fc100d07824b5d89437. Please customize variable to `lsp-ui-imenu-auto-refresh` for this feature! Thanks!

liebkne:
Does `lsp-ui-imenu-auto-refresh` care about switching between buffers?

Issue #295, opened 2019-06-11
lsp-ui-imenu autoupdate


vn-ki:
Add an option to autoupdate lsp-ui-imenu similar to the functionality offered by [imenu-list](https://github.com/bmag/imenu-list/#automatic-update).

This is very useful for navigating multiple big files.

yyoncho:
FYI I am working on a imenu gui control in lsp-treemacs which cover this functionality. (the code is there, I am for https://github.com/Alexander-Miller/treemacs/issues/287 be fixed).

axelson:
Any update on this? `lsp-ui-imenu` doesn't appear to update even if you close and re-open it.

jcs090218:
After commit 982b4669721c8bcd4dcf2fc100d07824b5d89437. Please customize variable to `lsp-ui-imenu-auto-refresh` for this feature! Thanks!

PR #515, opened 2020-09-21
Auto refresh imenu

jcs090218:
This PR is for feature #511 and #295. 

I have currently bind to only `after-save-hook`. Open for any other suggestion! Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Customization:
 
 - `lsp-ui-imenu-window-width` set window width
 - `lsp-ui-imenu--custom-mode-line-format` mode line format
+- `lsp-ui-menu-auto-refresh` auto refresh when necessary
 
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Customization:
 
 - `lsp-ui-imenu-window-width` set window width
 - `lsp-ui-imenu--custom-mode-line-format` mode line format
-- `lsp-ui-menu-auto-refresh` auto refresh when necessary
+- `lsp-ui-imenu-auto-refresh` auto refresh when necessary
 
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Customization:
 - `lsp-ui-imenu-window-width` set window width
 - `lsp-ui-imenu--custom-mode-line-format` mode line format
 - `lsp-ui-imenu-auto-refresh` auto refresh when necessary
-
+- `lsp-ui-imenu-refresh-delay` delay to refresh imenu
 
 ## Contributing
 
--- a/lsp-ui-imenu.el
+++ b/lsp-ui-imenu.el
@@ -65,7 +65,7 @@
   :type 'number
   :group 'lsp-ui-menu)
 
-(defcustom lsp-ui-menu-auto-refresh t
+(defcustom lsp-ui-imenu-auto-refresh t
   "Automatically refresh imenu when certain conditions meet."
   :type 'boolean
   :group 'lsp-ui-menu)
@@ -247,42 +247,44 @@ Return the updated COLOR-INDEX."
 (defun lsp-ui-imenu nil
   "Open ui-imenu in side window."
   (interactive)
-  (setq lsp-ui-imenu--origin (current-buffer))
-  (imenu--make-index-alist)
-  (let ((imenu-buffer (get-buffer-create lsp-ui-imenu-buffer-name))
-        (list imenu--index-alist))
-    (with-current-buffer imenu-buffer
-      (let* ((padding (lsp-ui-imenu--get-padding list))
-             (grouped-by-subs (-partition-by 'imenu--subalist-p list))
-             (color-index 0)
-             (bars (make-bool-vector lsp-ui-imenu--max-bars t))
-             (inhibit-read-only t))
-        (remove-overlays)
-        (erase-buffer)
-        (dolist (group grouped-by-subs)
-          (if (imenu--subalist-p (car group))
-              (setq color-index (lsp-ui-imenu--insert-items "" group padding bars 0 color-index))
-            (lsp-ui-imenu--put-separator)
-            (lsp-ui-imenu--insert-items "" group padding bars 1 color-index)
-            (setq color-index (1+ color-index))))
-        (lsp-ui-imenu-mode)
-        (when lsp-ui-imenu--custom-mode-line-format
-          (setq mode-line-format lsp-ui-imenu--custom-mode-line-format))
-        (goto-char (point-min))
-        (add-hook 'post-command-hook 'lsp-ui-imenu--post-command nil t)))
-    (let ((win (display-buffer-in-side-window imenu-buffer '((side . right)))))
-      (set-window-margins win 1)
-      (select-window win)
-      (set-window-start win 1)
-      (lsp-ui-imenu--move-to-name-beginning)
-      (set-window-dedicated-p win t)
-      ;; when `lsp-ui-imenu-window-width' is 0, fit window to buffer
-      (if (= lsp-ui-imenu-window-width 0)
-          (let ((fit-window-to-buffer-horizontally 'only))
-            (fit-window-to-buffer win)
-            (window-resize win 3 t))
-        (let ((x (- lsp-ui-imenu-window-width (window-width))))
-          (window-resize (selected-window) x t))))))
+  (lsp-ui-imenu-buffer-mode 1)
+  (let ((imenu-auto-rescan lsp-ui-imenu-auto-refresh))
+    (setq lsp-ui-imenu--origin (current-buffer))
+    (imenu--make-index-alist)
+    (let ((imenu-buffer (get-buffer-create lsp-ui-imenu-buffer-name))
+          (list imenu--index-alist))
+      (with-current-buffer imenu-buffer
+        (let* ((padding (lsp-ui-imenu--get-padding list))
+               (grouped-by-subs (-partition-by 'imenu--subalist-p list))
+               (color-index 0)
+               (bars (make-bool-vector lsp-ui-imenu--max-bars t))
+               (inhibit-read-only t))
+          (remove-overlays)
+          (erase-buffer)
+          (dolist (group grouped-by-subs)
+            (if (imenu--subalist-p (car group))
+                (setq color-index (lsp-ui-imenu--insert-items "" group padding bars 0 color-index))
+              (lsp-ui-imenu--put-separator)
+              (lsp-ui-imenu--insert-items "" group padding bars 1 color-index)
+              (setq color-index (1+ color-index))))
+          (lsp-ui-imenu-mode)
+          (when lsp-ui-imenu--custom-mode-line-format
+            (setq mode-line-format lsp-ui-imenu--custom-mode-line-format))
+          (goto-char (point-min))
+          (add-hook 'post-command-hook 'lsp-ui-imenu--post-command nil t)))
+      (let ((win (display-buffer-in-side-window imenu-buffer '((side . right)))))
+        (set-window-margins win 1)
+        (select-window win)
+        (set-window-start win 1)
+        (lsp-ui-imenu--move-to-name-beginning)
+        (set-window-dedicated-p win t)
+        ;; when `lsp-ui-imenu-window-width' is 0, fit window to buffer
+        (if (= lsp-ui-imenu-window-width 0)
+            (let ((fit-window-to-buffer-horizontally 'only))
+              (fit-window-to-buffer win)
+              (window-resize win 3 t))
+          (let ((x (- lsp-ui-imenu-window-width (window-width))))
+            (window-resize (selected-window) x t)))))))
 
 (defun lsp-ui-imenu--kill nil
   (interactive)
@@ -336,7 +338,25 @@ Return the updated COLOR-INDEX."
 (define-derived-mode lsp-ui-imenu-mode special-mode "lsp-ui-imenu"
   "Mode showing imenu entries.")
 
-(defun lsp-ui-imenu-enable (_enable))
+(defun lsp-ui-imenu--refresh ()
+  "Safe refresh imenu content."
+  (when (get-buffer lsp-ui-imenu-buffer-name)
+    (save-selected-window
+      (let ((lsp-ui-imenu-auto-refresh t))  ; Force refresh, ignore custom variable.
+        (lsp-ui-imenu)))))
+
+(defun lsp-ui-imenu-buffer--enable ()
+  "Enable `lsp-ui-imenu-buffer'."
+  (add-hook 'after-save-hook #'lsp-ui-imenu--refresh nil t))
+
+(defun lsp-ui-imenu-buffer--disable ()
+  "Disable `lsp-ui-imenu-buffer'."
+  (remove-hook 'after-save-hook #'lsp-ui-imenu--refresh t))
+
+(define-minor-mode lsp-ui-imenu-buffer-mode
+  "Minor mode 'lsp-ui-imenu-buffer-mode'."
+  :group lsp-ui-imenu
+  (if lsp-ui-imenu-buffer-mode (lsp-ui-imenu-buffer--enable) (lsp-ui-imenu-buffer--disable)))
 
 (provide 'lsp-ui-imenu)
 ;;; lsp-ui-imenu.el ends here
--- a/lsp-ui-imenu.el
+++ b/lsp-ui-imenu.el
@@ -92,7 +92,7 @@
 (declare-function imenu--subalist-p 'imenu)
 (defvar imenu--index-alist)
 
-(defvar lsp-ui-imenu--refresh-timer nil
+(defvar-local lsp-ui-imenu--refresh-timer nil
   "Auto refresh timer for imenu.")
 
 (defun lsp-ui-imenu--pad (s len bars depth color-index for-title is-last)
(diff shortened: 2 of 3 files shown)


brotzeit:
I think `after-save-hook` is fine for now, but @yyoncho suggested using a timer. IMO we should merge this and update the code when we have a more sophisticated solution.

jcs090218:
> I think `after-save-hook` is fine for now, but @yyoncho suggested using a timer. IMO we should merge this and update the code when we have a more sophisticated solution.

I was thinking about using a timer too. This patch is indeed a quick solution; I am looking forward to update this in the near future.

lsp-ui-imenu.el
```diff
@@ -331,7 +338,25 @@ Return the updated COLOR-INDEX."
 (define-derived-mode lsp-ui-imenu-mode special-mode "lsp-ui-imenu"
   "Mode showing imenu entries.")
 
-(defun lsp-ui-imenu-enable (_enable))
+(defun lsp-ui-imenu--refresh ()
```
brotzeit, line 341: I wonder if we should make this function a command and bind it to "r" in `lsp-ui-imenu-mode`.

jcs090218: Added!

jcs090218:
I have managed to use `timer` and `after-change-functions`. I also added `lsp-ui-util.el` file, but currently there aren't much yet. What do guys think about the `utility` module? I think is quite nice to have a `helper` file that holds all the generic functions/macros. Thanks!

brotzeit:
Tested it and seems to work correctly :+1:

brotzeit:
@jcs090218 do you know that you can close issues automatically https://docs.github.com/en/enterprise/2.13/user/articles/closing-issues-using-keywords ?

jcs090218:
> do you know that you can close issues automatically

Yeah, I think I knew it, but I like to do it separately. Maybe I am just not use to this feature yet.

@brotzeit I guess we can merge this now and improve in the next patch?

Thanks!

brotzeit:
I'd like to hear @sebastiencs opinion first.

README.md
```diff
@@ -108,6 +108,7 @@ Customization:
 
 - `lsp-ui-imenu-window-width` set window width
 - `lsp-ui-imenu--custom-mode-line-format` mode line format
+- `lsp-ui-menu-auto-refresh` auto refresh when necessary
```
brotzeit, line 112: `lsp-ui-imenu-refresh-delay` is missing

README.md
```diff
@@ -108,6 +108,7 @@ Customization:
...
```
brotzeit, line 111: menu -> imenu

jcs090218: Fixed!

README.md
jcs090218: Added! Thanks! 👍

sebastiencs:
Thanks for working on this @jcs090218 

> I think after-save-hook is fine for now, but @yyoncho suggested using a timer.

I'm not sure if using a timer on `after-change-functions` is the best idea, IMO the `lsp-ui-imenu` window is not really relevant while typing. And this could make quite a lot of requests, with potentially big buffers/slow servers.

What about refreshing it just on `after-save-hook` ? I think it's a more appropriate hook to use, and without the need of a timer

@yyoncho In what use cases a timer would be beneficial ?

jcs090218:
> I'm not sure if using a timer on `after-change-functions` is the best idea, IMO the `lsp-ui-imenu` window is not really relevant while typing. And this could make quite a lot of requests, with potentially big buffers/slow servers.

I think is quite nice to have a updated imenu while typing; of course, this kind of action would take some performance. Have you tried it yet? On my end it work quite well. I have managed to use `run-with-idle-timer` and set to the default delay to `1.0` seconds. It's quite hard to send a refresh request unless you type in a speed of 1 second per letter. Then yes, that's the worst case scenario. In average, I think the user can tweak the `lsp-ui-refresh-delay` base on their own typing habit.

> What about refreshing it just on `after-save-hook` ? I think it's a more appropriate hook to use, and without the need of a timer

I'm fine with this one too. Let me know if I need to revert my updates! Thanks!

brotzeit:
IMO you don't use imenu all the time, but only in certain situations and therefore it's not a big issue that the performance isn't optimal. However it would be nice if imenu would also update the buffer when you are focusing a different buffer than the initial buffer. I just realized that's currently not the case right ?

brotzeit:
@ulises what do you think ?

yyoncho:
> @yyoncho In what use cases a timer would be beneficial ?

this is what other editors do. Also, there is `lsp-on-change-hook` and we do load the symbols info for other components with no perf issues. lsp-treemacs-symbols has optimized rendering since it updates only the changed elements.

ulises:
Apologies for the delay in answering.

My use case would be to have lsp-symbols-imenu open side by side with a buffer that contains a class (or equivalent). So, I'd personally favour 2 hooks:
  - when the main buffer changes (you're working now on a different class or equivalent)
  - when the buffer is saved (you added/removed something in the current class or equivalent)

I'm not sure a timer makes sense considering that workflow, but perhaps I'm just too focused on my own needs :)

ulises:
A further comment, this time on behaviour though: I noticed that when you press `<RET>` on an imenu item the accompanying buffer scrolls to the symbol, but the focus/pointer remains on the imenu buffer. I suspect it'd be nicer if the pointer was placed in the other buffer (perhaps at the beginning of the symbol?).

wakira:
> the focus/pointer remains on the imenu buffer

@ulises Pressing `META-RET` should bring it to the main buffer. Have a try

jcs090218:
> So, I'd personally favour 2 hooks:

I think the current implementation should be good for the 2 use cases you mentioned by binding the `after-change-functions`. :)

brotzeit:
I would give it a try and see if people complain about performance issues.

brotzeit:
Would this be ok for you @sebastiencs ?

brotzeit:
Maybe it would also make sense to allow the user to choose if he wants to use a timer or not.

sebastiencs:
> Would this be ok for you @sebastiencs ?

Alright, let's do it with a timer.
Can we just make sure that the request is made asynchronously ? To not block while waiting for the server.
I don't remember how is made the request but if someone could check if it's already async, if not I think it needs to be implemented @jcs090218

jcs090218:
> Can we just make sure that the request is made asynchronously ? To not block while waiting for the server.
> 
> I don't remember how is made the request but if someone could check if it's already async, if not I think it needs to be implemented @jcs090218 

Any suggestion to make it async? As I have tested; I think the current implementation isn't async? Overall I am not sure if the execution is async/sync because the overall execution goes very fast and I did not feel the delay, yet I assumed is sync.

Right now I have implemented the refresh functionality by reusing the same function `lsp-UI-imenu`. I mean it would be great to have some suggestions from guy who created this function. @sebastiencs Is it you? :)

sebastiencs:
@jcs090218 I just check the code and it's not async.
`lsp-ui-imenu` calls `imenu--make-index-alist` which calls `lsp--imenu-create-index`.
`lsp--imenu-create-index` blocks until it receives a response from the server.

> the overall execution goes very fast and I did not feel the delay

That's true with your specific configuration and server. Some lsp servers might be slow and calling `lsp--imenu-create-index` in a timer will degrade the user experience.
It is not difficult to find on google comments about lsp-ui/mode being slow, so I prefer to be very careful with this kind of thing.

> > Maybe it would also make sense to allow the user to choose if he wants to use a timer or not.

> @brotzeit I think this is a good solution.

Since making it async will require some more work/refactoring, can we disable the timer by default, for this PR ? Until we find a solution to do it asynchronously.
If you want to work on that @jcs090218 I can help you.

> I would give it a try and see if people complain about performance issues.

I don't think it is the best strategy. A few users will complain in the issue tracker, most of them will just stop using lsp-ui.

jcs090218:
> I just check the code and it's not async.

Got you! Thanks for your confirmation! 👍 

> It is not difficult to find on google comments about lsp-ui/mode being slow, so I prefer to be very careful with this kind of thing.

Yes, of course. I agree with you.

> Since making it async will require some more work/refactoring, we can do it in another PR. Can we disable the timer by default, for this PR ? Until we find a solution to do it asynchronously.

Applied! @sebastiencs Let me know if this work! :)

lsp-ui-imenu.el
```diff
@@ -58,12 +59,27 @@
...
   :type 'number
-  :group 'lsp-ui-menu)
+  :group 'lsp-ui-imenu)
+
+(defcustom lsp-ui-imenu-auto-refresh t
+  "Automatically refresh imenu when certain conditions meet."
+  :type 'boolean
+  :group 'lsp-ui-imenu)
+
+(defcustom lsp-ui-imenu-refresh-delay 1.0
```
sebastiencs, line 74: If I understand correctly, this delay is used only when `lsp-ui-imenu-auto-refresh` is non-nil ?
If so, can you rename the variable to `lsp-ui-imenu-auto-refresh-delay`

lsp-ui-imenu.el
```diff
@@ -58,12 +59,27 @@
...
+  "Delay time to refresh imenu."
+  :type 'float
+  :group 'lsp-ui-imenu)
+
+(defcustom lsp-ui-imenu-refresh-after-change-p nil
```
sebastiencs, line 79: Please remove the suffix `-p`

lsp-ui-imenu.el
```diff
@@ -58,12 +59,27 @@
...
+  :group 'lsp-ui-imenu)
 
 (defcustom lsp-ui-imenu-window-width 0
   "When not 0, don't fit window to buffer and use value as window-width."
   :type 'number
-  :group 'lsp-ui-menu)
+  :group 'lsp-ui-imenu)
+
+(defcustom lsp-ui-imenu-auto-refresh t
+  "Automatically refresh imenu when certain conditions meet."
```
sebastiencs, line 70: Please set this to `nil` by default, as discussed this shouldn't be on by default. Until we make the request async.
And in the docstring please mention that it might slow down navigation/editing because the request is not async.

lsp-ui-imenu.el
```diff
@@ -331,7 +354,40 @@ Return the updated COLOR-INDEX."
...
+  "Safe refresh imenu content."
+  (interactive)
+  (let ((imenu-buffer (get-buffer lsp-ui-imenu-buffer-name)))
+    (when imenu-buffer
+      (save-selected-window
+        (if (equal (current-buffer) imenu-buffer)
+            (select-window (get-buffer-window lsp-ui-imenu--origin))
+          (setq lsp-ui-imenu--origin (current-buffer)))
+        ;; Force refresh, ignore custom variable.
+        (let ((lsp-ui-imenu-auto-refresh t)) (lsp-ui-imenu))))))
```
sebastiencs, line 367: This shouldn't call `lsp-ui-imenu`, `lsp-ui-imenu` does extra work to create a window.
You will need to split `lsp-ui-imenu`

lsp-ui-imenu.el
```diff
@@ -242,44 +261,47 @@ Return the updated COLOR-INDEX."
...
+            (if (imenu--subalist-p (car group))
+                (setq color-index (lsp-ui-imenu--insert-items "" group padding bars 0 color-index))
+              (lsp-ui-imenu--put-separator)
+              (lsp-ui-imenu--insert-items "" group padding bars 1 color-index)
+              (setq color-index (1+ color-index))))
+          (lsp-ui-imenu-mode)
+          (when lsp-ui-imenu--custom-mode-line-format
+            (setq mode-line-format lsp-ui-imenu--custom-mode-line-format))
+          (goto-char (point-min))
+          (add-hook 'post-command-hook 'lsp-ui-imenu--post-command nil t)))
```
sebastiencs, line 288: Only this part of code should be called by `lsp-ui-imenu--refresh`.
See the comment in `lsp-ui-imenu--refresh`

lsp-ui-util.el
```diff
@@ -0,0 +1,40 @@
...
+
+;;; Commentary:
+
+;;
+;; Utility module for Lsp-Ui.
+;;
+
+;;; Code:
+
+(defun lsp-ui-util-safe-kill-timer (tmr)
```
sebastiencs, line 35: Please rename this to `lsp-ui-util--safe-kill-timer`, it's a private function.
And `tmr` to `timer`

lsp-ui-imenu.el
```diff
@@ -76,6 +92,9 @@
 (declare-function imenu--subalist-p 'imenu)
 (defvar imenu--index-alist)
 
+(defvar lsp-ui-imenu--refresh-timer nil
```
sebastiencs, line 95: This should be a local variable, I think ?

lsp-ui-imenu.el
```diff
@@ -242,44 +261,47 @@ Return the updated COLOR-INDEX."
...
-      (lsp-ui-imenu--move-to-name-beginning)
-      (set-window-dedicated-p win t)
-      ;; when `lsp-ui-imenu-window-width' is 0, fit window to buffer
-      (if (= lsp-ui-imenu-window-width 0)
-          (let ((fit-window-to-buffer-horizontally 'only))
-            (fit-window-to-buffer win)
-            (window-resize win 3 t))
-        (let ((x (- lsp-ui-imenu-window-width (window-width))))
-          (window-resize (selected-window) x t))))))
+  (lsp-ui-imenu-buffer-mode 1)
```
sebastiencs, line 264: The mode is enabled here but it's not disabled anywhere

sebastiencs (changes requested):

lsp-ui-imenu.el
jcs090218: There is only one imenu window right? If this is true, the I don't see the benefit setting it to local.

lsp-ui-imenu.el
jcs090218: This is for `after-save`. I have temporary created another variable `lsp-ui-imenu-refresh-after-change` to disable automatically refresh imenu when `after-change`. So by default; the user can still have auto refresh when `after-save`, but not for the `after-change` because the variable `lsp-ui-imenu-refresh-after-change` is already set to `nil` by default. Let me know if you still want this to be disable by default.

lsp-ui-imenu.el
jcs090218: Changed! Thanks!

lsp-ui-imenu.el
jcs090218: Fixed! Thanks!

lsp-ui-imenu.el
jcs090218: Yes. I assumed `imenu` get activated when user calls `M-x lsp-ui-imenu` and not anywhere else. I didn't consider closing it because there is no exit point by the current design? 😕 We might have to discuss this too. Or maybe I have overlooked the code when I'm looking for the exit point.

lsp-ui-util.el
jcs090218: Fixed! Thanks!

lsp-ui-imenu.el
jcs090218: You meant the code from L265 to L288, right? Aren't the code 99% of the function `lsp-ui-imenu`? Any objection to just call the function? 😕

jcs090218: Okay, I may know why. `lsp-ui-menu` created the window and it constantly pop out.

brotzeit:
@jcs090218 can you please bump the version as the last commit.

lsp-ui-imenu.el
jcs090218: Okay, I have extracted the refresh execution to `lsp-ui-imenu--refresh-content` function.

jcs090218:
> can you please bump the version as the last commit.

I have bumped version to `6.3`. Does that work for you? Thanks!

lsp-ui-imenu.el
sebastiencs: hmm what about having just one variable `lsp-ui-imenu-auto-refresh` with 3 values:
- `t` for always
- `after-save` to ignore `after-change-functions`
- `nil` to disable

This will avoid having another variable l`sp-ui-imenu-refresh-after-change`

In all cases, both behaviors should be disabled by default, until we make it async

lsp-ui-imenu.el
sebastiencs: You can have multiple frames of emacs open, so there can be more than 1 imenu window

lsp-ui-imenu.el
sebastiencs: I think it should be disabled when the window of `lsp-ui-imenu` is closed

lsp-ui-imenu.el
jcs090218: Got you! Fixed! 👍

lsp-ui-imenu.el
jcs090218: See https://github.com/emacs-lsp/lsp-ui/pull/515/commits/1894bc88337576d5553b17449d7c17fc76ad8f37. Does this work for you?

lsp-ui-imenu.el
jcs090218: > hmm what about having just one variable lsp-ui-imenu-auto-refresh with 3 values:

Changed.

> In all cases, both behaviors should be disabled by default, until we make it async

Got you! I have make the variable default to `nil`.

brotzeit:
> I have bumped version to 6.3. Does that work for you? Thanks!

I mean just before we are merging this pull request, so every commit is included in the new version.

jcs090218:
@brotzeit Got it. Just want to confirm if I did that correctly. 😖

jcs090218:
Okay, I think I have implemented this the way @sebastiencs wants. Let me know if something you want me to change and I will fix it in the later patch! Merge this now!

Issue #483, opened 2020-08-07
too many cancel requests


dmakarov:
lsp-ui-sideline sends many cancel requests
https://github.com/emacs-lsp/lsp-ui/blob/e49780e41d33b73882453f25675324aa4bebc9d0/lsp-ui-sideline.el#L478
On pyls this has a dramatic effect of slowing down Emacs to a complete halt when actively editing a python source file.
Most of these requests are actually processed and warnings are generated for such cancelations that an unknown message id is requested to be canceled.  Can this performance problem be solved?

sebastiencs:
Fixed with https://github.com/emacs-lsp/lsp-ui/pull/490

PR #517, opened 2020-09-27
[lsp-ui-sideline] Reduce calls to line-number-at-pos

sebastiencs:
`line-number-at-pos` is the slowest operation in lsp-ui-sideline and was
called several times, including in post-command-hook.
It impacts performance only with big buffers.

This reduce the number of calls to this function to 0 or 1.
It is now called only after the timer `lsp-ui-sideline-delay`, and when the cursor changes of line.
When the cursor stays on the same line it doesn't call `line-number-at-pos`

@kurnevsky This should also fix the issue we mentioned in https://github.com/emacs-lsp/lsp-ui/pull/509
Can you please confirm ?

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -139,10 +139,18 @@ It is used to know when the cursor has changed its line or point.")
   "Value of window's width on the last operation.
 It is used to know when the window has changed of width.")
 
+(defvar-local lsp-ui-sideline--last-line-number nil
+  "Line number on the last operation.
+Used to avoid calling `line-number-at-pos' when we're on the same line.")
+
 (defvar-local lsp-ui-sideline--timer nil)
 
 (defvar-local lsp-ui-sideline--code-actions nil
   "Holds the latest code actions.")
+
+(defvar-local lsp-ui-sideline--cached-infos nil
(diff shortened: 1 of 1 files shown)


kurnevsky:
Yes, it fixes that issue.

kurnevsky:
Though I noticed one problem - if I move cursor quickly old requests are not cancelled and they start execute with new ones.

kurnevsky:
I fixed it in this commit: https://github.com/emacs-lsp/lsp-ui/commit/4094eaf557ae43229abf3a0cf6e53c0a446f7bf0
But for some reason it was deleted :(

sebastiencs:
> Though I noticed one problem - if I move cursor quickly old requests are not cancelled and they start to execute with new ones and redraw the same sideline multiple times.

Does this occurs on the master branch ?

> I fixed it in this commit: 4094eaf
But for some reason it was deleted :(

Ah, this has been removed for https://github.com/emacs-lsp/lsp-ui/issues/483

kurnevsky:
> Does this occurs on the master branch ?

Yes.

kurnevsky:
@sebastiansturm will you merge this? At least it doesn't make things worse :)

jcs090218 (approved):
I have tested and I think this should be good to merge! 💯 👍

Issue #519, opened 2020-09-30
lsp-ui-doc fails when cursor is at the top of window


brotzeit:
I could only reproduce this when the cursor was at a "}" sign which was at the top of the window. Maybe we should just use `ignore-errors` for this corner case since otherwise it works well, apparently.

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  lsp-ui-doc--mv-at-point(209 40 8 0)
  lsp-ui-doc--move-frame(#<frame  0x5558454c6068>)
  lsp-ui-doc--display(nil #("Receiver<Instant>" 0 17 (markdown-gfm-code (9 27) face (markdown-code-face) font-lock-multiline t)))
  lsp-ui-doc--callback(#<hash-table equal 2/2 0x15561263dbfd> (7919 . 7920) #<buffer channel.rs>)
  #f(compiled-function (hover) #<bytecode 0x56e4df09f44dc21>)(#<hash-table equal 2/2 0x15561263dbfd>)
  apply(#f(compiled-function (hover) #<bytecode 0x56e4df09f44dc21>) #<hash-table equal 2/2 0x15561263dbfd>)
  #f(compiled-function (&rest args) #<bytecode 0x25153383d487328>)(#<hash-table equal 2/2 0x15561263dbfd>)
  #f(compiled-function (result) #<bytecode 0x1294c140f0299451>)(#<hash-table equal 2/2 0x15561263dbfd>)
  #f(compiled-function (result) #<bytecode -0x73e6940c5cbe891>)(#<hash-table equal 2/2 0x15561263dbfd>)
[...]

PR #522, opened 2020-10-02
[lsp-ui-doc] Fix frame positioning when bound is outside window #519

sebastiencs:
@brotzeit This should fix your issue #519


brotzeit:
Seems to work, thanks!

PR #528, opened 2020-10-24
Remove incorrect documentation from lsp-ui-flycheck.el

Sodaware:
The opening documentation says `lsp-ui-flycheck-enable` should be called to enable lsp-ui-flycheck, but that function was removed.


danielmartin:
Thank you for your contribution!

PR #529, opened 2020-11-01
lsp-ui-doc-render-function made useful

ThChatz:
lsp-ui-doc-render-function was not being called when rendering
documentation. This patch aims to fix this by calling it from
lsp-ui-doc--callback


ThChatz:
after a deeper look, i think this needs some more work

PR #539, opened 2020-12-06
Update a couple faces to look good on light background's

JohnC32:
- lsp-ui-peek-highlight had handling for light background but dim gray is too light. Also
  can we add a red box as this looks good.

- lsp-ui-sideline-current-symbol had handling for light backgrounds, but order was wrong, also
  dim gray and the light background are too close, thus proposed a more visible colors

- lsp-ui-sideline-code-action had handling for light backgrounds, but order was wrong, fixed.

### DIFF
--- a/lsp-ui-peek.el
+++ b/lsp-ui-peek.el
@@ -109,13 +109,12 @@ By default, the peek view isn't shown if there is 1 xref."
   :group 'lsp-ui-peek)
 
 (defface lsp-ui-peek-highlight
-  '((((background light)) :background "dim gray"
-     :foreground "white"
-     :distant-foreground "black")
+  '((((background light)) :background "yellow"
+     :box (:line-width -1 :color "red"))
     (t :background "white"
        :foreground "black"
        :distant-foreground "white"
-       :box (:line-width -1 :color "white")))
+       :box (:line-width -1 :color "red")))
   "Face used to highlight the reference/definition.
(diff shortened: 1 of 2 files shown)


jcs090218:
LGTM, would you mind showing the before/after comparison screenshot? I just want to have a look so it's easier for everyone to see instead of the code! Thank you! ;)

JohnC32:
For lsp-ui-peek-highlight, here's the setup prior to my proposal

lsp-ui-peek-highlight-orig

and the proposal (notice the red box lets you see which getName is the reference on the line in addition to the highlighted symbol in yellow):

lsp-ui-peek-highlight-new

For lsp-ui-sideline-current-symbol, here's the setup prior to my proposal (notice the "[name]" below the square in the box on the right is invisible because it's white on white):

lsp-ui-sideline-current-symbol-old

and the proposal (notice the "[name]" below the square in the box is now black on white):

lsp-ui-sideline-current-symbol-new

For lsp-ui-sideline-code-action, notice the "fix-it" code action is hard to read because it's yellow on white:

lsp-ui-sideline-code-action-orig

and the proposal (notice that you can now read the fix-it code action because it's dark orange on white):

lsp-ui-sideline-code-action-new

Thanks
John

jcs090218 (approved):

jcs090218:
LGTM, Thanks!

Issue #540, opened 2020-12-09
Error CI


jcs090218:
I think there is something wrong with the CI. Is there something wrong due to the `actions/setup-python`? 😕

ericdallo:
Yeah, we just need to use 2.0 version instead of 1.1

ericdallo:
I only can open a PR on the weekend for that :/

jcs090218:
Close this now. Fixed by #541.

PR #541, opened 2020-12-10
Attempt to fix CI.

jcs090218:
For #540.


jcs090218:
@ericdallo It seems like there are other errors! 😕

ericdallo:
Yeah, probably removing this unnecessary echo should fix it: https://github.com/emacs-lsp/lsp-ui/blob/master/.github/workflows/test.yml#L53

jcs090218:
Thanks! It works!

Issue #489, opened 2020-08-27
Problem with display-line-number on emacs -nw or emacs-nox (no window)


apiraino:
Hi,

I use emacs 26.3 without graphical environment, have the Ubuntu `emacs-nox` package which basically corresponds to running `emacs -nw`. I've realized that `lsp-ui` is not looking great without GUI, the sideline is broken (while lsp-ui works perfectly on stardard graphical emacs). See here:

![recording](https://user-images.githubusercontent.com/6098822/91499046-1660db80-e8c1-11ea-96de-f4de033c3ce2.gif)

I'm running lsp-ui without any customization or `setq` parameter, just installed and activated it.

Is there any configuration I can try to fix this? Perhaps somewhere there is

thanks

sebastiencs:
@apiraino In terminals, `lsp-ui` can't use child frames, which is used in GUI.
In terminals the implementation is based on overlay.

> the sideline is broken

I think you're referring to `lsp-ui-doc` from you gif, but the gif is too fast I can't really see what/where is it broken.

In my terminal it renders like this:

![image](https://user-images.githubusercontent.com/5273820/91566263-f9d2ab00-e975-11ea-939e-e949de8d851a.png)

Do you have a different result with your config ?
What configuration would make your experience better ? Setting a maximum width and height ?

apiraino:
ah thanks a lot @sebastiencs for the hints and corrections! Now I'm starting to understand how it is supposed to work.

So, here is a more meaningful description of what happens. The doc overlay is too wide and each line of the documentation is cut in two lines (notice the `\` char at the end of each line).

I also suspect the documentation rendering is broken: what do I need to install to properly rendere the docs without using webkit?

The funny thing is that no matter my shell width, the overlay shows this behaviour. If I enlarge the shell, the overlay alignment gets fixed, but if I move back and forth triggering the doc overlay, it always appears broken until I resize the shell window. See this behaviour in this GIF (~5mb):

![recording](https://user-images.githubusercontent.com/6098822/91578885-173e4f80-e94b-11ea-933b-1ea206f8d492.gif)

I've tried fiddling with some settings, trying to force a different behaviour but I didn't see anything different
```
 '(lsp-ui-doc-alignment (quote window))
 '(lsp-ui-doc-border "red")
 '(lsp-ui-doc-header t)
 '(lsp-ui-doc-max-height 10)
 '(lsp-ui-doc-max-width 50)
[...]

PR #546, opened 2021-01-01
Adjust lsp-ui-doc overlay inner padding (when not using a childframe)

apiraino:
Hi,

this is an attempt to study the problem I've reported in #489. I've dug a bit into the code and I think the root of the issue is that `lsp-ui` has not a great support for emacsen that don't use a GUI (either running `emacs -nw` or without the GTK libs compiled in). Note: this is not meant to be a critique, just to frame the issue.

When not using the GUI childframe (i.e. `lsp-ui-doc-use-childframe nil`) it's not clear to me how the overlay is created and if I can interact with it. For example I cannot scroll a long page of documentation:

I understand the technical limits of a non-GUI overlay, I probably couldnt improve a lot here, but this PR is trying to fix at least one issue about the incorrect padding calculated. For some reason the overlay padding from the right border is not enough, therefore each line exceeding the overlay width "bleeds" into the main buffer messing the view (see my issue report).

I'm not converstant with elisp, so I'm very open to any explaination and suggestion to improve this patch.

Thanks for a review and for working on lsp-ui! :+1:


lsp-ui-doc.el
```diff
@@ -702,12 +702,18 @@ FN is the function to call on click."
 (defvar-local lsp-ui-doc--inline-width nil)
 
 (defun lsp-ui-doc--inline-window-width nil
+  "Calculate inner padding for the child frame showing the documentation"
+  (if (not (bound-and-true-p lsp-ui-doc-use-childframe))
+      (setq additional-width-spacing 8)
```
apiraino, line 707: the value of `8` here is a complete guess of what the minimum good value to fix the problem is. Same reasoning for the default value of `1` that lsp-ui-doc is already setting as an alternate value.

lsp-ui-doc.el
```diff
@@ -702,12 +702,18 @@ FN is the function to call on click."
...
+
   (- (min (window-text-width)
           (window-body-width))
      (if (bound-and-true-p display-line-numbers-mode)
          (+ 2 (line-number-display-width))
-       0)
-     1))
+         0)
+     additional-width-spacing)
+  )
```
jcs090218, line 716: Can you use `let` operator for local variable `additional-width-spacing`? And don't put `)` on a newline (L716). Thanks!

lsp-ui-doc.el
jcs090218: Make sense to me.

lsp-ui-doc.el
apiraino: done

jcs090218:
Not sure if you can use `let` operator this way. Let's try this.

```el
(defun lsp-ui-doc--inline-window-width nil
  "Calculate inner padding for the child frame showing the documentation"
  (let ((additional-width-spacing (if (bound-and-true-p lsp-ui-doc-use-childframe)
                                      8 1)))
    (- (min (window-text-width) (window-body-width))
       (if (bound-and-true-p display-line-numbers-mode)
           (+ 2 (line-number-display-width)) 0)
       additional-width-spacing)))
```

This is a much simplify version of function `lsp-ui-doc--inline-window-width`.

apiraino:
embarassing how little elisp I know :smile: 

and now the CI pass! thank you so much @jcs090218 for your help

jcs090218:
@apiraino No problem! Glad to help out!

Can someone test this patch? @sebastiencs @brotzeit

apiraino:
Unfortunately I think there are still issues with the overlay width. Simply setting a larger value (8 instead of 1) is not useful. I am testing with more code and for example the overlay breaks with long function signatures (example +100 characters).

I welcome a review and suggestions, but I'd keep this PR in standby until I figure out a better approach (example: by dynamically calculating the width).

jcs090218:
I might try to avoid magic number like `8` or `1`. Calculating the padding by longest text lines sounds like a good approach. ;)

apiraino:
I think I'll close this and clean up some mess I've left. I couldn't get around to fixing this and I can't reproduce it anymore after migrating to another Emacs config

PR #554, opened 2021-01-05
README.md: add two important lsp-ui-doc variables

rvl:
I found `lsp-ui-doc-show-with-cursor` to be super annoying! It took me ages to realise that it's possible to disable it. So I thought this variable is important enough to add to the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ Customization:
 - `lsp-ui-doc-enable` enable lsp-ui-doc
 - `lsp-ui-doc-position` Where to display the doc
 - `lsp-ui-doc-delay` Number of seconds before showing the doc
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
Nice catch! 

`lsp-ui` is really missing some good documentation, thank you for your contribution!

rvl:
Thanks @ericdallo - actually I found the documentation to be reasonably good.
Although perhaps it's not emacs newbie friendly.
Maybe I don't even know what I don't know about `lsp-ui`. :smile_cat:

ericdallo:
hahaha Yes, I hope someday other `emacs-lsp` pacakges docs would be as good as `lsp-mode` docs at least

PR #564, opened 2021-01-21
Update awesome-pages to new name

jcs090218:


ericdallo:
We probably should change it to use https://github.com/emacs-lsp/docs-image like lsp-mode and dap-mode

jcs090218:
> We probably should change it to use https://github.com/emacs-lsp/docs-image like lsp-mode and dap-mode

Sorry, I am still testing this doc deployment. Let me fix it.

PR #568, opened 2021-02-02
Fix miscalculation of first line

### LINKED ISSUE #567: lsp ui sideline not working in tsx buffer even though errors are highlighted.
**Describe the bug**

Reproducing the bug report from: https://github.com/hlissner/doom-emacs/issues/4598 since I think this is an lsp-mode issue.

lsp ui sideline not showing diagnostics in tsx buffer even though it is highlighted

The first line is supposed to show the following error in the sideline

```
1: typescript: 'React' is declared but its value is never read.
```

However it doesn't show. It is highlighted, and I can see it in the diagnostics buffer, but the sideline doesn't show up.

scrot_20210130_234804

However, if I disable `lsp-ui-sideline-mode` the error is shown right below below the cursor.

scrot_20210130_235235

**To Reproduce**
1. Create a typescript react app with `npx create-react-app client --template typescript`
2. Open the App.tsx file in src
3. Move your cursor to the first line to display the error
4. Disable `lsp-ui-sideline-mode` and try again.

**Expected behavior**
lsp ui sideline should show diagnostics in tsx buffer when I have my cursor on that line

**Which Language Server did you use**
ts-ls

**OS**
Linux x86_64

jcs090218:
This fix #567

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -189,10 +189,7 @@ This face have a low priority over the others."
 
 (defun lsp-ui-sideline--first-line-p (pos)
   "Return non-nil if POS is on the first line."
-  (save-excursion
-    (goto-char 1)
-    (forward-line)
-    (> (point) pos)))
+  (save-excursion (goto-char 1) (forward-line 1) (> (point) pos)))
 
 (defun lsp-ui-sideline--calc-space (win-width str-len index)
   "Calcul whether there is enough space on line.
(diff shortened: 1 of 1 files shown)


lsp-ui-sideline.el
```diff
@@ -224,10 +221,7 @@ if OFFSET is non-nil, it starts search OFFSET lines from user point line."
     (while (and (null pos) (<= (abs index) 30))
       (setq index (if up (1- index) (1+ index)))
       (setq pos (lsp-ui-sideline--calc-space win-width str-len index)))
-    (if (and up (or (null pos) (and (lsp-ui-sideline--first-line-p pos)
-                                    ;; line-end-position returns a wrong value when its
-                                    ;; argument lead to a line < 0, so we need to use this trick
-                                    (-any-p 'lsp-ui-sideline--first-line-p lsp-ui-sideline--occupied-lines))))
```
jcs090218, line 230: I don't think line ever returns lower than 0.

Issue #585, opened 2021-03-17
Weird character lsp-ui-doc while using lsp-grammarly


jcs090218:
I am getting the weird character ` ` while using `lsp-grammarly`. This doesn't happens in VSCode but in Emacs.

![Untitled-1](https://user-images.githubusercontent.com/8685505/111437110-aec78c00-873d-11eb-8982-b559d94cb846.png)

jcs090218:
Close this by #586.

PR #586, opened 2021-03-17
Fix weird character, #585

jcs090218:


lsp-ui-doc.el
```diff
@@ -917,7 +917,8 @@ BUFFER is the buffer where the request has been made."
            (thing-at-point 'symbol t)
            (-some->> contents
              lsp-ui-doc--extract
-             (replace-regexp-in-string "\r" ""))))
+             (replace-regexp-in-string "\r" "")
+             (replace-regexp-in-string " " " "))))
```
jcs090218, line 921: Interesting, the character ` ` cannot be seen on GitHub...

PR #588, opened 2021-03-17
Tweak document frame

jcs090218:


lsp-ui-doc.el
```diff
@@ -470,7 +464,7 @@ FRAME just below the symbol at point."
                          (if (< (car it) (window-start))
                              (cons 0 0)
                            (posn-x-y (posn-at-point (1- (window-end))))))))
-          (frame-relative-symbol-x (+ start-x x))
+          (frame-relative-symbol-x (+ start-x x (* (frame-char-width) 2)))
```
jcs090218, line 467: This move frame a bit to the left so you can see the whole frame.

PR #591, opened 2021-03-23
Fix by language specification

jcs090218:
Just realized the patch from #590 may not work for markdown content. Wrap it only when rendering is occurred!

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -297,7 +297,6 @@ Because some variables are buffer local.")
                                  (lsp:markup-content-kind marked-string)))
                         language))
           (markdown-hr-display-char nil))
-     (message "--> language: %s" language)
      (cond
       (lsp-ui-doc-use-webkit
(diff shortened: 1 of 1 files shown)


lsp-ui-doc.el
```diff
@@ -302,7 +302,10 @@ Because some variables are buffer local.")
        (if (and language (not (string= "text" language)))
            (format "```%s\n%s\n```" language string)
          string))
-      (t (lsp--render-string (lsp-ui-doc--inline-formatted-string string) language))))))
+      ;; For other programming languages
+      (language (lsp--render-string (lsp-ui-doc--inline-formatted-string string) language))
+      ;; For default markdown content
+      (t (lsp--render-element (lsp-ui-doc--inline-formatted-string string)))))))
```
jcs090218, line 308: Language is `nil` when the string content is in markdown. Render it with function `lsp--render-element` and use function `lsp--render-string` for the rest.

PR #592, opened 2021-03-23
Fix document string for marked string

jcs090218:


lsp-ui-doc.el
```diff
@@ -304,7 +304,7 @@ Because some variables are buffer local.")
          string))
       ;; For other programming languages
       (language (lsp--render-string (lsp-ui-doc--inline-formatted-string string) language))
-      ;; For default markdown content
+      ;; For default element content
```
jcs090218, line 307: I thought the default would be `markdown` content. I guess this would be more precise.

PR #618, opened 2021-05-21
Make the top and bottom margins be same height

### LINKED ISSUE #617: Different  heights of top and bottom margins
https://github.com/emacs-lsp/lsp-ui/blob/bd94da9348d9b4e814fdb55dae061826f48ffde9/lsp-ui-doc.el#L602
https://github.com/emacs-lsp/lsp-ui/blob/bd94da9348d9b4e814fdb55dae061826f48ffde9/lsp-ui-doc.el#L616

The top and bottom margins are different height. I think the height should be same. `0.3` is a good choice.

seagle0128:
Close #617 .


jcs090218 (approved):

jcs090218:
Thanks!

PR #621, opened 2021-06-04
[lsp-ui-doc] Prevent overriding face in lines below inline doc

gexplorer:
When displaying the doc **inline**, the lines below the documentation loose their format.

This is the code before displaying the inline doc:

without-doc

After displaying the inline doc, the face of the lines below it reset to default:

inline-doc

This PR fixes this problem:

inline-doc-after-PR

The problem is the way the the overlay is trying to prevent inheriting the face of the text below. After merging the **buffer-string** and the **doc-string** the `default` face was being applyed to the whole text. 

Instead of this the problem can be solved by applying the `default` face to the overlay.


jcs090218 (approved):
LGTM. Not sure why it's added in the first place? 😕

Issue #625, opened 2021-06-16
[feature request] Global toggle for lsp-ui-doc-enable


betaprior:
I would like to be able to toggle lsp-ui-doc popups, but using the `lsp-ui-doc-enable` variable I can only seem to control the state on emacs startup. For example, if I start with doc popups enabled but then do `(setq lsp-ui-doc-enable nil)`, it takes no effect.

The best I can do is toggle `lsp-ui-doc-mode`, but that only takes effect for a given buffer.  It would be nice to be able to have global toggles for lsp-ui features that can take effect across all buffers and not just on startup.

ericdallo:
I can't see an easy way to do that, any suggestion @yyoncho?

yyoncho:
> but using the `lsp-ui-doc-enable` variable I can only seem to control the state on emacs startup

I don't think that this is true. After reopening the buffer lsp-ui-doc should not be enabled(at least lsp-mode won't enable it).

betaprior:
> I don't think that this is true. After reopening the buffer lsp-ui-doc should not be enabled(at least lsp-mode won't enable it).

Ah, well I was thinking about more of an on-the-fly toggle that does not involve re-opening buffers. Some of us have problems with never killing buffers ;)

PR #626, opened 2021-06-17
Add global entry

jcs090218:
For #625.

Not sure if this is a good idea, but I have implemented first.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -1102,6 +1102,9 @@ If nil, do not prevent mouse on prefix keys.")
     (remove-hook 'post-command-hook 'lsp-ui-doc--make-request t)
     (remove-hook 'delete-frame-functions 'lsp-ui-doc--on-delete t))))
 
+(define-global-minor-mode global-lsp-ui-doc-mode lsp-ui-doc-mode
+  (lambda () (lsp-ui-doc-mode 1)))
+
 (defun lsp-ui-doc-enable (enable)
   "Enable/disable ‘lsp-ui-doc-mode’.
 It is supposed to be called from `lsp-ui--toggle'"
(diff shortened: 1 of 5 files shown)


ericdallo (approved):
LGTM but I don't know if there is any side effects with that, WDYT @yyoncho?

yyoncho (commented):
Don't we need something like this?

```
(defun lsp-ui-doc-disable (args)
  "..."
  (interactive "P")
  (setq lsp-ui-doc-enable nil)
  (mapc (lambda (b) (with-current-buffer b
                      (lsp-ui-doc-mode -1)))
        (buffer-list)))
```


for the enable we have to look for all buffers in lsp-mode state.

jcs090218:
> for the enable we have to look for all buffers in lsp-mode state.

No, I think the global minor mode will handles this? But the flag would not disable since it's `defcustom`.

```el
(define-global-minor-mode global-lsp-ui-doc-mode lsp-ui-doc-mode
  (lambda () (lsp-ui-doc-mode 1)))
```

yyoncho:
Ok, seems like the global mode will handle the turnoff. Don't we need also to set lsp-ui-doc-enable to nil to avoid starting ui-doc in the new buffers?

lsp-ui-doc.el
```diff
@@ -1110,9 +1113,10 @@ It is supposed to be called from `lsp-ui--toggle'"
 (defun lsp-ui-doc-show ()
   "Trigger display hover information popup."
   (interactive)
-  (lsp-ui-doc--callback (lsp-request "textDocument/hover" (lsp--text-document-position-params))
-                        (or (bounds-of-thing-at-point 'symbol) (cons (point) (1+ (point))))
-                        (current-buffer)))
+  (when lsp-ui-doc-mode
```
yyoncho, line 1116: this will prevent the user from using ui-doc without enabling lsp-ui-doc, right? I think this is the goal of lsp-ui-doc-show

jcs090218: I think it make sense it shows only when the `lsp-ui-doc` is enabled. So yes.

jcs090218:
> Ok, seems like the global mode will handle the turnoff. Don't we need also to set lsp-ui-doc-enable to nil to avoid starting ui-doc in the new buffers?

No, I think the minor will still enable if user have set it up in the config. For instance, 

```el
(global-lsp-ui-doc-mode 1)
```

Then it will automatically be turn on no matter what and I think that's the default action from Emacs. If user want to disable lsp-ui completely then it will have to evaluate expression manually.

```el
(setq lsp-ui-doc-enable nil)
```

lsp-ui-doc.el
yyoncho: The idea of lsp-ui-doc-show is to be used when automatic lsp-ui-doc is off. This change will allow showing the doc manually only when the automatic popup is enabled which does not make sense.

PR #627, opened 2021-06-17
Fix column calculation and displaying sideline in the first line

jcs090218:
This patch fix the following issue,

* Fix the sufficient space calculation
* Fix logic when displaying sideline in the first line


lsp-ui-sideline.el
```diff
@@ -574,7 +584,7 @@ Argument HEIGHT is an actual image height in pixel."
           (or (-some-> range (lsp-get :start) (lsp-get :line) (= line))
               (-some-> range (lsp-get :end) (lsp-get :line) (= line))))
         (lsp--get-buffer-diagnostics))
-       (apply 'vector)))
+    (apply 'vector)))
```
jcs090218, line 587: I think this is cause by indentation changes from `dash`?

Issue #613, opened 2021-05-19
Moving cursor is very laggy


danlamanna:
Related: https://github.com/emacs-lsp/lsp-mode/issues/837.

The lag goes away after running either `lsp-disconnect` or disabling `lsp-ui-mode`.

Emacs version:
```
GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.29, cairo version 1.17.4) of 2021-05-18
```

LSP info:
```
Checking for Native JSON support: OK
Check emacs supports `read-process-output-max': OK
Check `read-process-output-max' default has been changed from 4k: OK
Byte compiled against Native JSON (recompile lsp-mode if failing when Native JSON available): OK
`gc-cons-threshold' increased?: OK
Using gccemacs with emacs lisp native compilation (https://akrl.sdf.org/gccemacs.html): OK
```

The profiling report is:
```
                                                                         23  48% - lsp-ui-doc--make-request
                                                                         23  48%  - if
                                                                         23  48%   - progn
                                                                         23  48%    - let
                                                                         23  48%     - if
                                                                          9  19%      - if
                                                                          9  19%       - lsp-ui-doc--hide-frame
                                                                          9  19%        - if
                                                                          9  19%           progn
                                                                          8  17%      - lsp-ui-doc--hide-frame
                                                                          5  10%       - if
                                                                          5  10%        - progn
                                                                          4   8%         - if
                                                                          4   8%            let
                                                                          8  17% - timer-event-handler
                                                                          8  17%  - apply
                                                                          8  17%     #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_10>
                                                                          6  12%   winner-save-old-configurations
                                                                          4   8% - #<lambda 0x886c35a0f37e5a2>
                                                                          4   8%  - while
                                                                          4   8%   - if
                                                                          4   8%    - let*
                                                                          4   8%     - if
                                                                          4   8%      - progn
                                                                          4   8%       - lsp--parser-on-message
                                                                          4   8%        - condition-case
                                                                          4   8%         - let
                                                                          4   8%          - let*
                                                                          4   8%           - let*
                                                                          4   8%            - lsp--get-message-type
                                                                          4   8%             - if
                                                                          4   8%                if
                                                                          3   6% - redisplay_internal (C function)
                                                                          3   6%    mode-line-default-help-echo
                                                                          3   6% - sp--save-pre-command-state
[...]

PR #628, opened 2021-06-19
Delay post request to improve performance

jcs090218:
For #613

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -864,9 +864,7 @@ HEIGHT is the documentation number of lines."
 
 (defun lsp-ui-doc--make-request nil
   "Request the documentation to the LS."
-  (and (not track-mouse)
-       lsp-ui-doc-show-with-mouse
-       (setq-local track-mouse t))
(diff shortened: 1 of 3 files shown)


lsp-ui-doc.el
```diff
@@ -864,9 +865,7 @@ HEIGHT is the documentation number of lines."
 
 (defun lsp-ui-doc--make-request nil
   "Request the documentation to the LS."
-  (and (not track-mouse)
-       lsp-ui-doc-show-with-mouse
-       (setq-local track-mouse t))
+  (and (not track-mouse) lsp-ui-doc-show-with-mouse (setq-local track-mouse t))
```
jcs090218, line 868: I replace it with just one line so it's much cleaner IMO.

lsp-ui-util.el
```diff
@@ -32,7 +32,7 @@
 
 ;;; Code:
 
-(defun lsp-ui-util--safe-kill-timer (timer)
+(defun lsp-ui-util-safe-kill-timer (timer)
```
jcs090218, line 35: Just rename the util function from private to public.

PR #644, opened 2021-07-18
lsp-ui-doc--hide-frame: dont clear doc buffer on hide

kiennq:
We've already clear the doc buffer on show so no need to clear the doc buffer on hide, the `webkit` version is also not doing that.

Don't clear the doc buffer on hide provides another advantage when triggers `lsp-ui-doc--open-markdown-link` with the keyboard.
In that case, there's a race condition that `(point)` is calculated after the doc buffer has been cleared, thus there's no link to follow.
Not clear the doc buffer on hide resolve that.


jcs090218 (approved):
No objection on this.

kiennq:
@jcs090218 I don't have merge permission on this project so you will have to merge it for me though :D

ericdallo:
Just invited you @kiennq

PR #648, opened 2021-08-20
More robust way to locate the resource dir

sunlin7:
Hi,
There is an error in my local when active lsp-ui:
>Error running timer: (error "Can’t find library lsp-ui")

The root cause is lsp-ui will call `(find-library-name "lsp-ui")` but there is only `lsp-ui.elc` in my local (the `lsp-ui.el` is deleted for saving storage), then the error happened.
This patch will do better than the original way.
Please help review and merge this PR. 
Thanks

### DIFF
--- a/lsp-ui.el
+++ b/lsp-ui.el
@@ -38,7 +38,7 @@
 (require 'find-func)
 
 (defconst lsp-ui-resources-dir
-  (--> (find-library-name "lsp-ui")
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thanks!

sunlin7:
Hi Jen-Chieh, Thank you!

PR #669, opened 2021-12-07
Add additional info to lsp-ui-doc-enable docs

axelson:
I initially didn't expect `lsp-ui-doc-enable` to only be read when the buffer was started, I thought it would take effect for buffers that were already running.

Information about this only taking effect on buffer start is taken from https://github.com/emacs-lsp/lsp-ui/issues/625#issuecomment-863548628

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -63,7 +63,8 @@
   :link '(info-link "(lsp-ui-doc) Customizing"))
 
 (defcustom lsp-ui-doc-enable t
-  "Whether or not to enable lsp-ui-doc."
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Can you make our CI happy?

```
In toplevel form:
lsp-ui-doc.el:65:1: Error: custom-declare-variable `lsp-ui-doc-enable' docstring wider than 80 characters
make: *** [Makefile:25: unix-compile] Error 1
```

jcs090218:
Let's merge this since the byte-compile will ignore such a warning for now!

PR #673, opened 2021-12-29
Prevent reference count string from jumping into the next line because of the window divider's width not taken into account

anonimitoraf:
Note that the leeway is hardcoded to 1 because that's just how it is in
TTY. TTY ignores window-divider-width

Before the code change:
image

After the code change:
image


ericdallo (approved):
LGTM

anonimitoraf:
Interesting, are those failures expected?^

PR #674, opened 2022-01-02
Organize, clean up, and improve performance

jcs090218:
This patch does the following,

- Move `lsp-ui-util` to `lsp-ui`
- Clean up some code
- Improve performance with macro `lsp-ui--with-no-redisplay`

PR #676, opened 2022-01-03
Revert from #674

jcs090218:
Let me revert the changes first. Not sure why the forward declaration does not work. 😕


bk138:
Thanks. #674 more or less completely broke emacs-lsp for me.

Issue #675, opened 2022-01-03
Error while checking syntax automatically: (invalid-function lsp-ui--with-no-redisplay)


liebkne:
When typing in a buffer with `lsp-mode`, got error `Error while checking syntax automatically: (invalid-function lsp-ui--with-no-redisplay)` with the latest `lsp-ui`.

Settings: 

```
(use-package lsp-ui
  :custom-face
  (lsp-ui-doc-background ((t (:background "gray95"))))
  (lsp-ui-peek-filename ((t (:inherit success))))
  (lsp-ui-peek-footer ((t (:inherit lsp-ui-peek-header))))
  (lsp-ui-peek-header ((t (:inherit default :background "gray95" :overline t :underline "black"))))
  (lsp-ui-peek-highlight ((t (:background "#e99ce8" :weight bold))))
  (lsp-ui-peek-list ((t (:inherit default :background "gray98"))))
  (lsp-ui-peek-peek ((t (:inherit default :background "gray98"))))
  (lsp-ui-peek-selection ((t (:background "#1a4b77" :foreground "white"))))
  (lsp-ui-sideline-code-action ((t (:inherit success))))

  :hook (lsp-mode . lsp-ui-mode)
  :bind (:map lsp-ui-mode-map ("M-." . lsp-ui-peek-find-definitions))
  :init
  (setq lsp-ui-sideline-ignore-duplicate t
        lsp-ui-sideline-show-hover nil
        lsp-ui-sideline-show-symbol nil
        lsp-ui-doc-delay 1.0
        lsp-ui-doc-enable nil
        lsp-ui-doc-alignment 'window))
```

emacs version: `GNU Emacs 27.2 (build 1, x86_64-apple-darwin21.1.0, Carbon Version 165 AppKit 2113) of 2021-11-18`.

ericdallo:
@jcs090218 https://github.com/emacs-lsp/lsp-ui/pull/674

shassard:
I'm also running into the same issue on Linux with Emacs 27.2 pulling the latest lsp-ui from melpa.

jcs090218:
I'm on it!

shassard:
FWIW: It seems that the latest melpa snapshot has this fixed now.

ericdallo:
We can close this issue as the PR was reverted or wait for the revert of revert get merged right @jcs090218 ?

jcs090218:
Yeah, we can close this.

PR #677, opened 2022-01-03
Organize, clean up, and improve performance (Continue #674)

jcs090218:
This should avoid error #675, but we will need to keep `lsp-ui-util.el` file.


jcs090218:
Hmm... CI seems to fail in Emacs version 26.3.

PR #687, opened 2022-03-12
Unified CI using Cask for windows

jcs090218:
We should able to use Cask on Windows, so unified the CI workflows on all platforms.


ericdallo (approved):
finally!

ericdallo:
it seems to be failing yet 🤷🏻

jcs090218:
Yeah, work in progress. I think I may did something wrong, but let me work on it a bit more. This may not be merged though. :0

Issue #695, opened 2022-04-06
`lsp-ui-doc-glance` auto-hide is not triggering


Lenbok:
The documentation for this function says that the documentation will automatically hide on next typing, however this does not happen, in my experience.

Perhaps my reading of the code is wrong, but it seems the mechanism for clearing is unreliable, in that the clearing of the doc is triggered via `post-command-hook` but the retrieval of the documentation from the language server happens asynchronously (i.e. it's quite likely that the documentation is returned *after* `lsp-ui-doc--glance-hide-frame` has  been removed from `post-command-hook`, which would lead to the documentation persisting). 

I also put a `message` at the start of `lsp-ui-doc--glance-hide-frame` and it seems to be called *immediately*, rather than after the *next* command. Does adding a function to post-command-hook while a command is *currently* executing result in that function being called when the current command finishes executing? If so, this also seems like the wrong behaviour.

It seems that the addition of `lsp-ui-doc--glance-hide-frame` to `post-command-hook` should instead happen once the documentation is returned from the server.

Edit: the bug only affects inline doc display mode.

jcs090218:
I agree with you. Even though I rarely see the issue from the current implementation! 😅 

Thanks for the changes!

jcs090218:
I think this broke `lsp-ui-doc-show` command. I have reverted #696 with #710 for now.

Let me reopen this issue.

PR #696, opened 2022-04-06
Move triggering of lsp-ui-doc auto-hide to after display

Lenbok:
This is my attempt to address #695.

As well fixing auto-hide for docs explicitly invoked via `lsp-ui-doc-glance`, this adds post-command clearing to both `lsp-ui-doc-show-with-mouse` and `lsp-ui-doc-show-with-cursor`. That seems consistent to me, so I don't think it presents a problem.

- Maybe there should be a separate `lsp-ui-doc-auto-hide` setting to disable auto-hiding for all three of these mechanisms?

Fixes: #695

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -958,7 +958,10 @@ BUFFER is the buffer where the request has been made."
            (-some->> contents
              lsp-ui-doc--extract
              (replace-regexp-in-string "\r" "")
-             (replace-regexp-in-string " " " "))))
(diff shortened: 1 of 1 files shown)


Lenbok:
@yyoncho @ericdallo Would you be able to review this fix?

Lenbok:
(rebased to master)

Lenbok:
@yyoncho @jcs090218 Is someone able to review / merge this?

jcs090218:
Thanks!

PR #700, opened 2022-04-11
Use Eask for CI

jcs090218:
I have recently created a tool named [Eask](https://github.com/emacs-eask/eask). It's similar to Cask, but it has a better cross-platform capability. It tests your package/configuration in a much more consistent way.

This PR does the following:

1. Replace Cask with Eask
2. Test Emacs 28.1
3. Remove `test/windows-bootstrap.el`
4. `test.yml` now has only one job (merged unix + windows)


ericdallo:
Amazing @jcs090218! Makes sense to me

ericdallo (approved):

ericdallo:
I have lsp-dart CI broken for windows for a while, it may be a good idea to replace with that

PR #701, opened 2022-04-18
Removes incorrect org-style markers from code block.

flamingbear:
simple typo correction.  The page is rendering with `~` marks.


jcs090218 (approved):

jcs090218:
Thanks!

Issue #654, opened 2021-09-22
Tab-bar-mode causes tab-bars to appear in childframes for lsp-ui info


RKBK:
If using childframes to display e.g. signature info at point, while at the same time having activated tab-bar-mode, the childframe will also get a tab-bar. This childframe tab-bar has no use, and will cause the height of the childframe to change, obscuring the point. 

To reproduce:
1. Enter tab-bar-mode, e.g. by opening a new tab (C-x t n)
2. Open a buffer where lsp and lsp-ui are running, having set lsp-ui to use childframes (and not e.g. xwidgets)
3. Symbol at point info will look something like this: 
![image](https://user-images.githubusercontent.com/4726978/134414226-8cc3d68c-0e5e-4777-b0e5-13ff14f6e972.png)

PR #704, opened 2022-04-25
suppress tab-bar for child-frame of lsp-ui-doc

buzztaiki:
This PR will fixes https://github.com/emacs-lsp/lsp-ui/issues/654


buzztaiki:
`tab-bar-lines-keep-state` is needed in Emacs28.
It might be better to set `tab-bar-format` to `nil` with buffer local, as in corfu.

https://github.com/minad/corfu/commit/f3e26eeec42f8bd293ffb698333fa48f8635b4f5

jcs090218 (approved):

jcs090218:
LGTM, thanks!

PR #711, opened 2022-05-15
Fix lsp-ui-doc-glance auto-hide

### LINKED ISSUE #695: `lsp-ui-doc-glance` auto-hide is not triggering

Lenbok:
Fixes #695

For some inexplicable reason, `lsp-ui-doc-show` calls `lsp-ui-doc--callback` synchronously, while `lsp-ui-doc-glance` went via a more tortuous asynchronous route (via `lsp-ui-doc--make-request`).

This commit makes `lsp-ui-doc-glance` equivalent to calling `lsp-ui-doc-show` followed by *delayed* addition to `post-command-hook` (since we're already in a command) of the `lsp-ui-doc--glance-hide-frame` function.

Note that `lsp-ui-doc--glance-hide-frame` is also simplified to just remove itself from `post-command-hook` and directly call `lsp-ui-doc-hide`.


kiennq:
> For some inexplicable reason, `lsp-ui-doc-show` calls `lsp-ui-doc--callback` synchronously, while `lsp-ui-doc-glance` went via a more tortuous asynchronous route (via `lsp-ui-doc--make-request`).

The async route is actually recommended since the sync call will basically hang Emacs if the request takes too much time.

Can you test with `(setq lsp-ui-doc-show-with-cursor nil)`, trigger `lsp-ui-doc-glance`, focus on the doc frame then focus back to the document to see if the doc frame is dismissed automatically?

lsp-ui-doc.el
```diff
@@ -1164,25 +1164,18 @@ It is supposed to be called from `lsp-ui--toggle'"
   (interactive)
   (lsp-ui-doc--hide-frame))
 
-(defvar-local lsp-ui-doc--unfocus-frame-timer nil)
 (defun lsp-ui-doc--glance-hide-frame ()
-  "Hook to hide hover information popup for `lsp-ui-doc-glance'."
-  (when (or (overlayp lsp-ui-doc--inline-ov)
```
kiennq, line 1170: > Note that `lsp-ui-doc--glance-hide-frame` is also simplified to just remove itself from `post-command-hook` and directly call `lsp-ui-doc-hide`.

This is done purposely to make sure that when the doc frame is focused, we can scroll inside the doc frame (the `post-command-hook` is triggered) and when we back out to the document, the `lsp-ui-doc--hide-frame` will be triggered.
The condition here is to ignore the calling to hide frame (and remove this function from `post-command-hook`) when the doc frame is visible

kiennq (changes requested):

Lenbok:
> The async route is recommended since the sync call will hang Emacs if the request takes too much time.

OK, so rather than make `lsp-ui-doc-glance` synchronous, we should make `lsp-ui-doc-show` asynchronous?  I had thought that the async aspect may have been to do with the mouse and cursor activation being triggered after delay, whereas `lsp-ui-doc-show` (and `lsp-ui-doc-glance`) are triggered by direct user interaction.

> Can you test with (setq lsp-ui-doc-show-with-cursor nil), trigger lsp-ui-doc-glance, focus on the doc frame then focus back to the document to see if the doc frame is dismissed automatically?

I use TTY emacs, where neither childframe nor webkit work. Does the overlay version even have the concept of focus? (I can't work out how to set the focus there if it does)

lsp-ui-doc.el
Lenbok: I can understand not wanting to auto-close it when the doc frame is focused, but what has that got to do with visibility? That sounds like it could be the cause of the original bug (where the doc window triggered via glance was never auto-hiding).

kiennq: Sorry I misread the condition. The condition here actually is "Don't call and unhook from `post-command-hook` yet when the doc frame is not visible (which can take time to set up)".

That's also the root cause of the original bug, which happens for overlay. Currently, we only check if the overlay is there, not checking its visibility. Thus the `lsp-ui-doc--hide-frame` will be called when the overlay is not even visible yet.
The simplest change is just to add a new condition there.
``` el
  (when (or (and (overlayp lsp-ui-doc--inline-ov) (overlay-buffer lsp-ui-doc--inline-ov))
```

You may want to spin it into a new function mimicking the frame version, `lsp-ui-doc--overlay-visible-p`)

kiennq:
> > The async route is recommended since the sync call will hang Emacs if the request takes too much time.
> 
> OK, so rather than make `lsp-ui-doc-glance` synchronous, we should make `lsp-ui-doc-show` asynchronous? I had thought that the async aspect may have been to do with the mouse and cursor activation being triggered after delay, whereas `lsp-ui-doc-show` (and `lsp-ui-doc-glance`) are triggered by direct user interaction.

I think they serve different purpose with glance as a asynchronous version of `lsp-ui-doc-show` that less punishing even when the user invokes it by themselves.

> > Can you test with (setq lsp-ui-doc-show-with-cursor nil), trigger lsp-ui-doc-glance, focus on the doc frame then focus back to the document to see if the doc frame is dismissed automatically?
> 
> I use TTY emacs, where neither childframe nor webkit work. Does the overlay version even have the concept of focus? (I can't work out how to set the focus there if it does)

The overlay seems doesn't support focus so I'm not sure how to test on that as well. Can you test with GUI Emacs instead?

Lenbok:
> I think they serve different purpose with glance as a asynchronous version of `lsp-ui-doc-show` that less punishing even when the user invokes it by themselves.

I do not agree -- the only difference AFAIK between `lsp-ui-doc-show` and `lsp-ui-doc-glance` is that the latter will auto-dismiss after next (out of focus) command? Since they're both explicitly initiated by the user they have the same level of "punishing". It just makes the code harder to follow if they take different paths for no clear reason.

I have experimented some more using the code at current master (370022b), putting some message statements into when `lsp-ui-doc--glance-hide-frame` is called, and also when the code makes it past the `overlayp` / `lsp-ui-doc--frame-visible-p` condition...

When using inline display (TTY), the first time glance is invoked upon starting emacs, it auto hides when expected (although the hide frame function is called immediately, it does not proceed past the condition). But from then on, it triggers auto hide **immediately** (when `post-command-hook` runs), before the async documentation retrieval has even returned the doc content! When the async retrieval comes back, the documentation then shows and never hides (since the hook has already been cleared by then). This is the behaviour I observed when I reported #695

When using childframe (GUI), the auto hiding seems to work every time. 

This suggests that the condition in:
https://github.com/emacs-lsp/lsp-ui/blob/master/lsp-ui-doc.el#L1170-L1171
is not working for inline docs. Looking now it seems due to `lsp-ui-doc--inline-ov` not being reset in `lsp-ui-doc--hide-frame`. So a minimal fix to #695 is to address that.


However, in terms of testing the focus behaviour with childframe (still on master), I'm not sure how you are normally supposed to get focus into the docs from `lsp-ui-doc-glance` though, as moving the mouse triggers the window to hide. If I try `M-x` the window hides. If I directly bind a key to `lsp-ui-doc-focus` and try that after bringing up the docs with glance, the docs hide. So it seems to me master is already broken in terms of handling doc focus from `lsp-ui-doc-glance`.

I did find another couple of bugs when using childframe (they might need testing in a minimal emacs environment):
- if the mouse pointer happens to be in the area where the child frame is to appear, `lsp-ui-doc-glance` never shows the doc at all.
- when you bring up the docs via `lsp-ui-doc-show`, then set focus to the docs, then `M-x` the UI locks up.

lsp-ui-doc.el
Lenbok: No, it turns out it works the first time. The root cause is that the overlay variable isn't being cleared when the overlay itself is deleted. I've updated the PR.

kiennq: I think we purposely don't reset the overlay that way so that a new overlay doesn't have to be created every time the inline doc is hidden.
The condition checking `(overlay-buffer lsp-ui-doc--inline-ov)` returns the buffer that the overlay is in, if it becomes `nil`, means the overlay has been deleted, which is similar to what you do if you clear the overlay. The good thing is that we don't have to create a new overlay

kiennq:
> I do not agree -- the only difference AFAIK between `lsp-ui-doc-show` and `lsp-ui-doc-glance` is that the latter will auto-dismiss after next (out of focus) command? Since they're both explicitly initiated by the user they have the same level of "punishing". It just makes the code harder to follow if they take different paths for no clear reason.

I think the difference here is that the `lsp-ui-doc-glance` is async while `lsp-ui-doc-show` is locking Emacs.
Locking up editor on any occasion is never a good thing so I'm in favor of the current implementation for `lsp-ui-doc-glance`, which follows the one that runs when you set `lsp-ui-doc-show-with-cursor` to `t`.


> Looking now it seems due to `lsp-ui-doc--inline-ov` not being reset in `lsp-ui-doc--hide-frame`. So a minimal fix to #695 is to address that.

The new fix is to reset the overlay, which can be a perf problem since we have to recreate a new overlay every time.
Is the proposed fix not working on your end? It does working in my test env though.



> However, in terms of testing the focus behaviour with childframe (still on master), I'm not sure how you are normally supposed to get focus into the docs from `lsp-ui-doc-glance` though, as moving the mouse triggers the window to hide. If I try `M-x` the window hides. If I directly bind a key to `lsp-ui-doc-focus` and try that after bringing up the docs with glance, the docs hide. So it seems to me master is already broken in terms of handling doc focus from `lsp-ui-doc-glance`.
> 
> I did find another couple of bugs when using childframe (they might need testing in a minimal emacs environment):
> 
> * if the mouse pointer happens to be in the area where the child frame is to appear, `lsp-ui-doc-glance` never shows the doc at all.
> * when you bring up the docs via `lsp-ui-doc-show`, then set focus to the docs, then `M-x` the UI locks up.

I bind `lsp-ui-doc-focus-frame` to focus onto the child frame.
Your experience is very different from mine, what's your env?
614a55c2-f6c0-4c45-95a4-3ebbc960f137

The problem with the mouse pointer in the area where the child frame is to appear happens for me as well, I guess that's one of the quirks of mixing mouse + keyboard input with child frame.
We do have PR for switching to use `posframe` but it seems not yet merged. @ericdallo 
You can use the mouse to trigger the docs to show and then move to focus onto the doc frame, the code path is very similar to what `lsp-ui-doc-glance` do, both use `lsp-ui-doc--make-request`

ericdallo:
@kiennq yeah, that PR is really old and stale, but it still makes sense, feel free to re-try from master

Lenbok:
@kiennq 

> The new fix is to reset the overlay, which can be a perf problem since we have to recreate a new overlay every time.
> Is the proposed fix not working on your end? It does working in my test env though.

I had actually missed your proposed fix before I made mine. I think it's speculation as to whether there is any performance difference either way (moving an overlay may be more expensive in some cases than creating a new one), but took your suggestion anyway, as it allowed a couple of other small tidy-ups.

I also went back to part of my first version in making `lsp-ui-doc-glance` go via `lsp-ui-doc-show` -- this is because the original version would only display after `lsp-ui-doc-delay`, which doesn't make any sense. For safety and clarity I also ensured the hide function is only added to `post-command-hook` **after** the current command finishes.

lsp-ui-doc.el
```diff
@@ -1178,11 +1180,12 @@ It is supposed to be called from `lsp-ui--toggle'"
 (defun lsp-ui-doc-glance ()
   "Trigger display hover information popup and hide it on next typing."
   (interactive)
-  (let ((lsp-ui-doc-show-with-cursor t))
-    (lsp-ui-doc--make-request))
+  (lsp-ui-doc-show)
   (when lsp-ui-doc--unfocus-frame-timer
     (cancel-timer lsp-ui-doc--unfocus-frame-timer))
-  (add-hook 'post-command-hook 'lsp-ui-doc--glance-hide-frame))
+  (run-at-time 1 nil    ;; Since we want hiding after *next* command, not *this* command
```
kiennq, line 1186: The `run-at-time` is scheduling the lambda on the next main loop already. Can we make this `0` instead of `1`, so the hide will be triggered immediately?

kiennq:
> I also went back to part of my first version in making `lsp-ui-doc-glance` go via `lsp-ui-doc-show` -- this is because the original version would only display after `lsp-ui-doc-delay`, which doesn't make any sense. For safety and clarity (and correctness if called when the doc is already showing) I also ensured the hide function is only added to `post-command-hook` **after** the current command finishes.

Thank you @Lenbok for working on addressing the comments.

Can we bind the `lsp-ui-doc-delay` to `0` instead of going to `lsp-ui-doc-show`?
Same argument as before, hanging Emacs while waiting for a glance document seems not good behavior.

lsp-ui-doc.el
Lenbok: Done.

Lenbok:
> Can we bind the `lsp-ui-doc-delay` to `0` instead of going to `lsp-ui-doc-show`?
> Same argument as before, hanging Emacs while waiting for a glance document seems not good behavior.

I really don't like that idea. I think it is significantly clearer with the code the way it is here (just look at all the shenanigans `lsp-ui-doc--make-request` does that is completely irrelevant to `lsp-ui-doc-glance` and `lsp-ui-doc-show`), and it's not any worse than what `lsp-ui-doc-show` is already doing.

If you want to make `lsp-ui-doc-show` async, perhaps that could be a separate PR.

Lenbok:
Let's reword that. They are both asynchronous under the hood (that's what `lsp-request` actually does) so they are interruptible rather than "hanging emacs" - it's just that they both show the documentation immediately as part of the command. Which is exactly what the user would expect for **both** of those commands. So it's not a regression, it's an improvement.

(Whereas mouse and cursor hovers can appear while the user is generally moving/mousing around and is likely to continue moving/mousing. It's a totally different use case and makes sense for those to be a different code path).

kiennq:
> Let's reword that. They are both asynchronous under the hood (that's what `lsp-request` actually does) so they are interruptible rather than "hanging emacs" - it's just that they both show the documentation immediately as part of the command. Which is exactly what the user would expect for **both** of those commands.

`lsp-request` is waiting for results from LSP server, while blocking Emacs from inputting and redisplay, overall, it makes Emacs hang. You can interrupt it with `Ctrl+G` but it may cause the internal not properly cleaned up.
If you want the sync version that can be interrupted by input, `lsp-request-while-no-input` is the way to go.

On the other hand, the part about showing documentation immediately can be solved by binding `lsp-ui-doc-delay` to `0` when calling `lsp-ui-doc--make-request`. It doesn't block Emacs input and redisplay, that means even with language server takes a long time to return the docs, Emacs will not go into hang state where you can't move cursor, switch windows. etc...

Lenbok:
> `lsp-request` is waiting for results from LSP server, while blocking Emacs from inputting and redisplay, overall, it makes Emacs hang. You can interrupt it with `Ctrl+G` but it may cause the internal not properly cleaned up.
> If you want the sync version that can be interrupted by input, `lsp-request-while-no-input` is the way to go.

We don't want it to be interrupted by input - if the user has initiated either of these two functions they expect to get the result back right then and aren't planning on doing anything else in the interim. In my experience showing the doc takes no appreciable time. If they change their mind, there is `C-g`. If the lsp server takes too long, there's already a timeout. You're basically trying to argue that no interactive function in the whole lsp-mode/lsp-ui code base should directly call `lsp-request`. That seems like nonsense to me, and looking at the code base it's also not true, I can see dozens of instances that follow this pattern.

kiennq:
> We don't want it to be interrupted by input - if the user has initiated either of these two functions they expect to get the result back right then and aren't planning on doing anything else in the interim

I don't agree with this.
And actually, that is not related to the problem you want to fix in this bug.

If you think the new behavior is better, please make a new PR for that while keeping the minimal fix for this problem.
That would be better to move the discussion and have this ready for merging.

Lenbok:
Separate PR created from the first commit on this branch.

Lenbok:
@kiennq Please re-review this PR. 

PR description at the top has been updated. This new strategy addresses several interrelated issues, meaning it now makes sense to keep in a single PR. Previous review comments are addressed:
- All doc showing is asynchronous
- Overlays are re-used. 

(It does not address the other possible performance issue we discovered related to inline + `display-line-numbers-mode`, the impact was not always present in our benchmarking, only small on a per-doc basis anyway, and can easily be done separately if desired)

I have tested inline and childframe docs every way I can think of and have not found anything undesirable.

Lenbok:
@kiennq would you (or anyone else) be able to review?

Lenbok:
Anyone able to review?

kiennq:
This PR seems to break the auto dismiss of the doc frame after it has been focused in (for scrolling and reading on the docs) and out.
I've bond `tab` to `lsp-ui-doc-focus-frame` and `q` to `lsp-ui-doc-unfocus-frame`.
Here is the current behavior
79ab2167-e8cf-48b2-9455-1470c43be3aa

As you can see, it's broken and I have to press Tab twice for the focus to be moved to the doc frame.
@Lenbok  @brotzeit

brotzeit:
Sorry...

That's the reason why I'm so hesitant to merge pull requests. I'm always missing something when I'm testing changes.

Issue #184, opened 2018-09-30
Sideline overflows current line with bigger font size


Infinisil:
When I increase the font size with `C-x C-+`, the sideline will sometimes overflow the line onto the next one:

![screenshot_3](https://user-images.githubusercontent.com/20525370/46262016-ac383400-c4fb-11e8-9cff-addf1597fcce.png)

Very annoying, because when moving over such lines while coding, half of the screen gets thrown around because of the additional line(s).

yyoncho:
Should be fixed with https://github.com/emacs-lsp/lsp-ui/pull/475

Infinisil:
Even with latest master it doesn't appear to be fully fixed. While it now takes more than just one `C-x C-+`, it still overflows after a bunch:
![2020-07-24_18-45](https://user-images.githubusercontent.com/20525370/88415292-853ca780-cdde-11ea-820d-e09098d924e7.png)

yyoncho:
Can you try turning off line numbers?

Infinisil:
Oh that apparently makes it worse. I was using `global-display-line-numbers-mode` for reference. Turning it off, with the standard font size:
![2020-07-24_20-29](https://user-images.githubusercontent.com/20525370/88423750-9096cf80-cdec-11ea-91e1-90f9edd4b43e.png)

Then doing `C-x C-+` once:
![2020-07-24_20-29_1](https://user-images.githubusercontent.com/20525370/88423774-9e4c5500-cdec-11ea-9402-5a42daa45568.png)

---

In comparison, here is the same sequence with `global-display-line-numbers-mode` turned on: Standard font size:
![2020-07-24_20-31](https://user-images.githubusercontent.com/20525370/88423886-d8b5f200-cdec-11ea-9501-9300b2b4b3b5.png)

Note how the text isn't flush on the right side anymore (maybe a clue as to what's going wrong)

Then doing `C-x C-+` once:
![2020-07-24_20-32](https://user-images.githubusercontent.com/20525370/88423905-df446980-cdec-11ea-8901-0031a99e83e9.png)

Note how the text doesn't overflow yet (but it does with more `C-x C-+`), but it got a bit closer to the righthand edge.

sebastiencs:
@Infinisil It should be fixed with https://github.com/emacs-lsp/lsp-ui/commit/3686ed86ac37c74bfdf2e05f96c968d082b826e7

Infinisil:
Can confirm, thanks!

[...]

PR #719, opened 2022-06-19
fix(align): align according to text-scale

jcs090218:
For #184 and #716

### DIFF
--- a/lsp-ui-sideline.el
+++ b/lsp-ui-sideline.el
@@ -299,13 +299,13 @@ MARKED-STRING is the string returned by `lsp-ui-sideline--extract-info'."
   (+ (apply '+ lengths)
      (if (display-graphic-p) 1 2)))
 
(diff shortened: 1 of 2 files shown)


lsp-ui-sideline.el
```diff
@@ -299,13 +299,13 @@ MARKED-STRING is the string returned by `lsp-ui-sideline--extract-info'."
...
      (if (display-graphic-p) 1 2)))
 
-(defun lsp-ui-sideline--compute-height nil
+(defun lsp-ui-sideline--compute-height ()
   "Return a fixed size for text in sideline."
   (if (null text-scale-mode-remapping)
       '(height 1)
     ;; Readjust height when text-scale-mode is used
     (list 'height
-          (/ 1 (or (plist-get (cdr text-scale-mode-remapping) :height)
```
jcs090218, line 308: This will always return `nil`. Can someone confirm this? 😕

Issue #714, opened 2022-05-27
Incorrect imenu display for some Rust enums


gcohara:
With the following example code:

```
pub enum MyEnum {
    First{
        hiya: usize
    },
    Second,
    Third,
}

fn main() {
    println!("Hello, world!");
}

mod foo {
    struct bar {
        hello: usize
    }
}

```

`lsp-ui-imenu-mode` looks like this:

<img width="141" alt="Screenshot 2022-05-27 at 13 54 31" src="https://user-images.githubusercontent.com/50339401/170703124-4d2f4801-cad2-40b9-91fe-8e854b335966.png">

I would expect the `First` variant to be below `MyEnum`, similar with `bar` and `foo`.
Is it possible that imenu is struggling with double nested stuff?

I'm using `lsp-ui` version 8.0.0, Emacs 27.2, and `rust-analyzer` 2022-05-18 release

brotzeit:
This has to be fixed in the rust-mode function `rust-imenu-generic-expression`. Maybe it's possible to create the imenu index with lsp. The rust-mode code is very old.

Issue #527, opened 2020-10-23
lsp-ui-imenu broken


rileyrg:
using the repo prepared for issue [Cant set break point in top level js code of react app.](https://github.com/emacs-lsp/dap-mode/issues/388)  lsp-ui-imenu is broken. Here handleClick is in the wrong place. Latest versions with emacs 27 from source.

Actual source if this is enough to reproduce:-

[index.js](https://github.com/rileyrg/dap-mode-react/blob/master/src/index.js)

![Screenshot from 2020-10-23 16-18-26](https://user-images.githubusercontent.com/362891/97036899-ccb30b00-1568-11eb-91ee-c3fb422b82b5.png)
i

minikN:
I think I experience the same
![image](https://user-images.githubusercontent.com/1888338/97708776-9b32c600-1ab9-11eb-9ee1-09b536f8293a.png)

jcs090218:
Is this issue related to #511? Or just somehow the imenu is broken when you startup? 😕

minikN:
> Is this issue related to #511? Or just somehow the imenu is broken when you startup?

No it's been like that before I started fiddling with auto refresh.

jcs090218:
> No it's been like that before I started fiddling with auto refresh.

Got it! Thanks for the clarification! 👍

czikus:
I'm seeing a similar problem in org-mode

brotzeit:
Should be fixed.

Issue #474, opened 2020-07-11
Wrong order of single-branch imenu tree items


tarao:
## Problem

For example, suppose we are editing the following Scala file with lsp-metals.

```scala
package some
package example
package namespace

object Main {
  def main(args: Array[String]): Unit =
    println("Hello, world!")
}
```

The expected result of `M-x lsp-ui-imenu` would be the following:

![image](https://user-images.githubusercontent.com/132569/87224963-5c6ee800-c3c4-11ea-9527-dce6d1f60616.png)

But we get this:

![image](https://user-images.githubusercontent.com/132569/87225037-eae36980-c3c4-11ea-90de-1e51c0634a0e.png)

## Reason

I found that `overlay-put`ing `after-string`s at the same point will show up in the reverse order.

```elisp
(progn
  (insert "\n")
  (let ((ov (make-overlay (point) (point))))
    (overlay-put ov 'after-string "one"))
  (let ((ov (make-overlay (point) (point))))
    (overlay-put ov 'after-string "two"))
  (let ((ov (make-overlay (point) (point))))
    (overlay-put ov 'after-string "three"))) ; C-x C-e here
threetwoone
```

If we have `priority`s, then the order is correct.

```elisp
(progn
  (insert "\n")
  (let ((ov (make-overlay (point) (point))))
    (overlay-put ov 'after-string "one")
    (overlay-put ov 'priority -1))
  (let ((ov (make-overlay (point) (point))))
    (overlay-put ov 'after-string "two")
    (overlay-put ov 'priority -2))
  (let ((ov (make-overlay (point) (point))))
    (overlay-put ov 'after-string "three")
    (overlay-put ov 'priority -3))) ; C-x C-e here
onetwothree
```

## Possible patch

I confirmed that just putting monotonically decreasing `priority`s at every overlay with an `after-string` will fix the problem.  Should I make a pull request with this?

```diff
diff --git a/lsp-ui-imenu.el b/lsp-ui-imenu.el
index 92ef0bb..b24430f 100644
--- a/lsp-ui-imenu.el
+++ b/lsp-ui-imenu.el
@@ -164,10 +164,13 @@
     (goto-char (+ (* depth 3) (line-beginning-position) padding))))

 (defvar lsp-ui-imenu--origin nil)
+(defvar lsp-ui-imenu--overlay-priority 0)

 (defun lsp-ui-imenu--put-separator nil
   (let ((ov (make-overlay (point) (point))))
[...]

PR #721, opened 2022-06-20
Fix `lsp-ui-imenu` sub-title ordering.

DivvyCr:
The `lsp-ui-imenu` displayed items in the wrong hierarchy, which was previously found in Issues [#474](https://github.com/emacs-lsp/lsp-ui/issues/474), [#527](https://github.com/emacs-lsp/lsp-ui/issues/527) [#632](https://github.com/emacs-lsp/lsp-ui/issues/632), and [#714](https://github.com/emacs-lsp/lsp-ui/issues/714), and in Java from my experience:

I believe I found a quick fix, which arose due to multiple elisp overlays being used in a recursive function.

In any case, this is the result after my quick fix:

(Also, the multi-line change in the pull-request is just me removing a redundant `let` block)

### DIFF
--- a/lsp-ui-imenu.el
+++ b/lsp-ui-imenu.el
@@ -202,7 +202,7 @@
   (let ((ov (make-overlay (point) (point)))
         (title-color (lsp-ui-imenu--get-color (+ color-index depth))))
     (overlay-put
-     ov 'after-string
+     ov 'before-string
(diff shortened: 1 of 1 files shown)


brotzeit:
I've tried it and it seems to work. Let's give it a try, thanks!

Issue #543, opened 2020-12-25
Lock lsp-ui-imenu window width


indigoviolet:
I wanted to exclude the lsp-ui imenu window from `balance-windows` etc, so I tried adding a hook to `lsp-ui-imenu-mode-hook` with `(window-preserve (get-buffer-window lsp-ui-imenu-buffer-name) t t)`, but that didn't work. I had to create a new function like

```
(defun my/lsp-ui-imenu ()
 (interactive)
 (lsp-ui-imenu)
 (window-preserve (get-buffer-window lsp-ui-imenu-buffer-name) t t)
)
```

Is this expected? Can it be fixed or made configurable?

yyoncho:
given the nature of emacs, imho creating such functions when the user wants a particular behaviour is acceptable.

indigoviolet:
@yyoncho Of course, but my question was really about the hook: why do you think that didn't work? Is it because of when the `lsp-ui-imenu-mode` is set up, and should that be delayed to the last minute so that the hook would work as expected?

jcs090218:
Sorry for the late reply.

Are you using any plugin that may effect the window configuration when you start `lsp-ui-imenu`? I don't see the balancing behaviour with the default `lsp-ui`'s imenu module. 😕

indigoviolet:
to be clear, i want to exclude the imenu window from being balanced when i
call balance-windows myself, it isn't happening automatically. my question
is about why the imenu mode hook is being fired in a way that i can't call
window-preserve in the hook.

jcs090218:
Can you tell me where you bind for balance-windows? The hook should work because the hook itself is called after window is created. 😕

DivvyCr:
I think I'm currently looking into this exact issue - I want to implement some of the treemacs behaviour into the imenu window, which includes making it 'invisible' during `balance-windows` (alongside making it persistent under `delete-other-windows`). Hence, my current 'fix' is identical to treemacs' implementation, relying on `window-size-fixed`:

Set `window-size-fixed` in the imenu buffer via the major-mode:
```diff
(define-derived-mode lsp-ui-imenu-mode special-mode "lsp-ui-imenu"
  "Mode showing imenu entries."
+  ;; The `lsp-ui-imenu' may need to be refreshed to take effect:
+  (setq window-size-fixed (if lsp-ui-imenu-persistent-buffer 'width nil)))
```

[...]

PR #722, opened 2022-07-01
Add an option that will impose `window-size-fixed` on `imenu` buffer.

DivvyCr:
Based on the discussion of Issue [#543](https://github.com/emacs-lsp/lsp-ui/issues/543).

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ Customization:
 - `lsp-ui-imenu-kind-position` place to show entries kind
 - `lsp-ui-imenu-buffer-position` place to show the buffer window
 - `lsp-ui-imenu-window-width` set window width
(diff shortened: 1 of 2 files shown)


DivvyCr:
I would suggest, however, to make this functionality to work with `lsp-ui-imenu-window-width`. I believe the expected behaviour of that variable imples that the `imenu` window will be effectively fixed.

Also, on a somewhat unrelated note, I considered adding the `no-delete-other-windows` property to the `imenu` window, which basically keeps the window on `C-x 1` invocation. Again, I feel that this could be part of the expected behaviour (in the 'persist' mode).

brotzeit:
Can you please add a note https://github.com/emacs-lsp/lsp-ui#lsp-ui-imenu

brotzeit:
Thanks!

PR #727, opened 2022-08-07
Fix lsp-ui-doc-glance not auto-hide after focus out from doc frame

### LINKED ISSUE #726: lsp-ui-doc frame is not hidden after the doc frame is unfocused
Seems to be regressed by #711, see https://github.com/emacs-lsp/lsp-ui/pull/711#issuecomment-1207260887

> This PR seems to break the auto dismiss of the doc frame after it has been focused in (for scrolling and reading on the docs) and out.
I've bond `tab` to `lsp-ui-doc-focus-frame` and `q` to `lsp-ui-doc-unfocus-frame`.
Here is the current behavior
79ab2167-e8cf-48b2-9455-1470c43be3aa
>
> As you can see, it's broken and I have to press Tab twice for the focus to be moved to the doc frame.

kiennq:
Fix #726


Lenbok:
This does not work for me, the window hides as soon as I attempt to set focus into the window.

kiennq:
> This does not work for me, the window hides as soon as I attempt to set focus into the window.

Are you doing that in overlay instead of doc frame? AFAIK, the overlay doesn't support focus.
For doc frame, you have to move your mouse out of where the doc frame is going to show up else the frame will be dismissed immediately (a limitation of child frame, same reason why `posframe` has to banish mouse to the top-left corner when its child frame is shown up).

Lenbok:
While I normally use overlay, I am testing with child frame. The position of the mouse makes no difference to me, this still does not work.

kiennq:
> While I normally use overlay, I am testing with child frame. The position of the mouse makes no difference to me, this still does not work.

Can you take some screencast?
I'm using both Ubuntu/Windows and this change is working fine.

PR #734, opened 2022-09-24
Use lsp-feature? to check for hover capability instead of lsp--capability

theschmocker:
lsp--capability doesn't correctly pickup hover capability when using omnisharp. lsp-feature? does. This fix lsp-ui-doc when using omnisharp

Related: https://github.com/emacs-lsp/lsp-mode/issues/3735


brotzeit:
:+1:

PR #743, opened 2023-01-16
Fix `peek--get-references` for `lsp-use-plists`

erikarvstedt:
This fixes `lsp-ui-peek-find-definitions/references` when `lsp-use-plists` is `t` and a single item is returned by the server.

#### Copy of commit msg
When `lsp-use-plists` is t, `lsp-request` returns a single plist (of form `(:key val ...)` or a list of plists.

Previously, when `lsp-use-plists` was t and a single plist was returned, the plist wasn't wrapped into a list, causing the subsequent code to fail.

### DIFF
--- a/lsp-ui-peek.el
+++ b/lsp-ui-peek.el
@@ -726,7 +726,10 @@ references.  The function returns a list of `ls-xref-item'."
 Returns item(s)."
   (-when-let* ((locs (lsp-request method params))
                (locs (if (listp locs)
-                         locs
+                         (if (symbolp (car locs))
+                             ;; A single plist was returned
+                             (list locs)
+                           locs)
                        (if (vectorp locs)
                            (append locs nil)
                          (list locs)))))


erikarvstedt:
@yyoncho, can `lsp-request` ever return a vector for methods "textDocument/references" or "textDocument/definition"?
If no, we can simplify the fixed fn.

yyoncho:
@erikarvstedt the solution is usually to use `seq` methods to handle the cases when it is unclear whether the result is list or vector.

erikarvstedt:
Using `seq` doesn't help in this specific case.

Which JSON objects in server responses are deserialised as vectors?
`textDocument/{references,definition}` will probably never return vectors as the top-level object, right?

yyoncho:
@erikarvstedt vector is returned for json vectors when using native json deserialization. When it is not used - you get list.

erikarvstedt:
Alright, then this PR can be merged.

yyoncho:
> Alright, then this PR can be merged.

In general lsp-use-plists should not be used outside of lsp-protocol.el

erikarvstedt:
Fixed.

yyoncho (approved):

PR #764, opened 2024-02-18
fit lsp-ui-imenu to longest line in buffer

LemonBreezes:
Previously, in buffers with many nested levels of imenu, the long lines would trail off past the width of the window. Now, we properly enlarge the window to fit the longest line.

### DIFF
--- a/lsp-ui-imenu.el
+++ b/lsp-ui-imenu.el
@@ -323,12 +323,22 @@ ITEMS are used when the kind position is `left."
 	  (let ((window-size-fixed)) ;; Temporarily set `window-size-fixed' to nil for resizing.
 	;; When `lsp-ui-imenu-window-width' is 0, fit window to buffer:
 	(if (= lsp-ui-imenu-window-width 0)
-			(let ((fit-window-to-buffer-horizontally 'only))
-			  (fit-window-to-buffer win)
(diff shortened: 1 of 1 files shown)


LemonBreezes:
Hello @jcs090218, when you get a chance, can you please review this PR? This PR fixes the window width not being adjusted to the size of the longest imenu entry. Before, imenu entries would trail off past the edge of the window and now they no longer do that for me (unless they make extensive use of the `display` property of overlays).

jcs090218:
I don't know if this is a good design since there is a chance the imenu window will take up all the space. 🤔

LemonBreezes:
> I don't know if this is a good design since there is a chance the imenu window will take up all the space. 🤔

I mean, this only applies when the `lsp-ui-imenu-window-width` is 0. Basically, what I am saying is that the current behavior for auto-resizing the window to the buffer does not work.

LemonBreezes:
Presumably it's because `fit-window-to-buffer` does not actually use the longest line in the entire buffer. The documentation for it says: "The new width of WINDOW is calculated from the maximum length of its buffer's lines that follow the current start position of WINDOW."

jcs090218 (approved):

jcs090218:
It makes sense, then. Thank you!

PR #775, opened 2024-04-23
fix(ui-doc): Hide doc frame on next post command

jcs090218:


brownts:
Hi @jcs090218, I'm seeing some odd behavior that seems to be related to this commit.  In the past, I've been able to move the mouse pointer over the child frame and use the mouse wheel to scroll the child frame.  Now, it appears as soon as I move the mouse pointer, the child frame disappears, so I can't scroll the child frame at all anymore.    This worked either with `lsp-ui-doc-show-with-cursor` or with `lsp-ui-doc-show-with-mouse`.  Can this capability be restored?

Another oddity appears to be the alternation of the child frame being displayed or hidden.  I believe this happens both with mouse and cursor hover settings (i.e., `lsp-ui-doc-show-with-mouse` and `lsp-ui-doc-show-with-cursor` respectively), but is most prominent with the cursor hover setting.  In that configuration, it's easy to see the problem.  If point is over a word that causes the child frame to be displayed, each time I move the point (for example starting at the left of the word and moving right one character at a time), the child frame will alternate between being displayed and not being displayed.  In the past, the child frame would stay displayed the entire time (possibly repositioning if `lsp-ui-doc-position` was set to `at-point`) as the point was moved to the right.

jcs090218:
I'll take a look!

brownts:
Looks like someone else is having a similar problem related to this change:
- https://www.reddit.com/r/emacs/comments/1cn6o7i/can_no_longer_click_on_doc_link_in_lsp_hover/

hellified:
Yep, that's my cry for help on Reddit. I'll watch here for further developments. Super annoying to lose the ability to click the link in the box and pop open the browser with full docs.

I can volunteer to be a guinea pig for changes if that will help.

jcs090218:
I've reverted the changes!

PR #779, opened 2024-06-01
Properly provide and require test-helper

manphiz:
* This ensures that tests have access to the helper functions defined in test/test-helper.el.

### DIFF
--- a/test/lsp-ui-flycheck-test.el
+++ b/test/lsp-ui-flycheck-test.el
@@ -1,6 +1,7 @@
 ;; -*- lexical-binding: t -*-
 
 (require 'lsp-modeline)
+(require 'test-helper)
 
 (ert-deftest lsp-ui-test-flycheck-list--update ()
   "Test if `lsp-ui-flycheck-list--update' populates buffer *lsp-diagnostics*."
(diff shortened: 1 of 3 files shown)


jcs090218:
`test-helper.el` is loaded by default using `ert-runner`. 🤔

manphiz:
Ah, this is mostly useful for distribution that doesn't have/use eask, e.g. Debian.  We have a custom script that directly run ert tests, in which case adding a proper provide/require will help.

jcs090218:
> Ah, this is mostly useful for distribution that doesn't have/use eask, e.g. Debian. We have a custom script that directly run ert tests, in which case adding a proper provide/require will help.

See https://github.com/jcs-elpa/reveal-in-folder/issues/3#issue-784460807.

manphiz:
AIUI, `test/test-helper.el` is not a test file per se, but a library/helper for tests, so I don't think it should be prevented to use provide.  For other actual test files (with `ert-deftest`) the suggestion apply, of course.

jcs090218:
This test is intended to be used with ert-runner, and since `test-helper.el` will be loaded by default, I’m sorry, but I don't see the need to merge this. 🤔 😅

manphiz:
Fair enough :)  It's OK.  I will carry a patch to cope with the restrictions in Debian.  Thanks!

jcs090218:
Thank you for understanding! :)

PR #780, opened 2024-06-01
Fix lsp-ui-doc--make-smaller-empty-lines

manphiz:
* This fixes the test with the current code.
* Tested on Emacs 29.3.


jcs090218:
Sorry for the late reply. I wonder what this PR solves? 🤔

manphiz:
Well, as mentioned in the first post, the tests are failing without the patch.

PR #784, opened 2024-11-30
[fix] `lsp-ui` not displaying pop-up on MarkedString[]

niontrix:
Why?:
- There seems to be a typo in extracting the documentation from an array of MarkedString[]. If `lsp-ui-doc-include-signature` is disabled, we still want to display the documentation in the pop-up which should reside in the rest of the MarkedString[] array excluding the first entry.

This change addresses the need by:
- Turn `cadr` into `cdar` to extract the rest of the first list inside `groups`

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -351,7 +351,7 @@ Because some variables are buffer local.")
                             (append list-marked-string nil))))
     (if lsp-ui-doc-include-signature
         list-marked-string
-      (cadr groups))))
+      (cdar groups))))
 
 (defun lsp-ui-doc--extract (contents)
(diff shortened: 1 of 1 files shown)


niontrix:
@jcs090218 could you please take a look and merge it in or provide feedback?

jcs090218 (approved):

jcs090218:
I think it makes sense. Can you provide screenshots before and after, just for the record? 🤔 Thanks!

niontrix:
@jcs090218 here are the before and after screenshots:

before:
Screenshot_20241204_185845

after:
Screenshot_20241204_185952

jcs090218:
Thank you!

PR #785, opened 2025-01-13
fix(doc): lsp-ui-sideline-diagnostic-max-lines

raxod502:
Was experiencing https://github.com/emacs-lsp/lsp-ui/issues/150 and had to look through a number of issues before finding this discussion which explained the user option which can be customized to change the behavior. Document.


jcs090218 (approved):

jcs090218:
Thank you!

Issue #795, opened 2025-10-30
Mouse movement invokes commands and lsp-ui-doc popup disappears


rytswd:
I'm using a separate frame for lsp-ui-doc, but with this issue, I cannot focus the frame, nor scroll any of the popup, as any mouse movement would clear out the doc popup.
It is also related to https://github.com/emacs-lsp/lsp-ui/issues/532, where any mouse movement cancels active isearch.

The installation is nothing special, here is my use-package definition

```elisp
  (use-package lsp-ui
    :after lsp-mode
    :commands lsp-ui-mode
    :custom
    (lsp-ui-sideline-enable nil)
    (lsp-ui-imenu-auto-refresh 'after-save)
    (lsp-ui-doc-position 'bottom)
    (lsp-ui-doc-side 'right)
    (lsp-ui-doc-show-with-cursor nil)
    (lsp-ui-doc-show-with-mouse nil)
    (lsp-ui-doc-max-height 25)
    (lsp-ui-doc-max-width 60)
    :hook
    ((lsp-ui-imenu-mode . (lambda ()
                            (hide-mode-line-mode 1)
                            ))
     )
    )
```

This seems to be both caused by the same `track-mouse` handling, and the simplest way to fix in a buffer was to `(setq-local track-mouse nil)`. I got a potential fix with AI generated solution, which may give us a pointer for how this can be handled correctly for various use cases. At least the fix I've got works for me, so I'm opening a PR with that.

PR #798, opened 2026-01-16
Fix lsp-ui-doc-mode non-idempotency

radon-at-beeper:
Per https://www.gnu.org/software/emacs/manual/html_node/elisp/Minor-Mode-Conventions.html:

> Enabling or disabling a minor mode twice in direct succession should not fail and should do the same thing as enabling or disabling it only once. In other words, the minor mode command should be idempotent. 

The existing implementation of `lsp-ui-doc-mode` was not idempotent. The first time it was enabled, it ran `lsp-ui-doc--setup-mouse`, which invoked:

```elisp
(setq lsp-ui-doc--mouse-tracked-by-us (not track-mouse))
(setq-local track-mouse t)
```

The second time it was enabled, it did the same thing, but now `track-mouse` was already non-nil, making `lsp-ui-doc--mouse-tracked-by-us` nil, even though `lsp-ui-doc` was in fact tracking the mouse. As a consequence, `lsp-ui-doc--disable-mouse-on-prefix` stopped functioning, causing a bug similar to https://github.com/emacs-lsp/lsp-ui/issues/795, where if you typed a prefix key such as `C-x` and then moved the mouse (in a buffer where lsp-ui-doc-mode was enabled) it would abort the key sequence.

You might not expect this sequence of events to come up commonly, but the default configuration of lsp-mode (or at least the one running in my config) apparently invokes `(lsp-ui-doc-mode 1)` three separate times when setting up a new buffer. I don't know why. But this caused the prefix bug to trigger universally for me.

I resolved the issue by adding a second mode variable for tracking whether or not `lsp-ui-doc--setup-mouse` needs to be run again. There may be additional idempotency issues, but this resolves one of them, at least.

### DIFF
--- a/lsp-ui-doc.el
+++ b/lsp-ui-doc.el
@@ -1170,6 +1170,11 @@ If nil, do not prevent mouse on prefix keys.")
 (defun lsp-ui-doc--prevent-focus-doc (e)
   (not (frame-parameter (cadr e) 'lsp-ui-doc--no-focus)))
 
+(defvar lsp-ui-doc-mode--was-already-enabled nil
+  "Whether the minor mode was previously enabled.
+This is used to notice if the mode is enabled again when it was already
+enabled, to avoid re-copying previous state.")
+
 (define-minor-mode lsp-ui-doc-mode
   "Minor mode for showing hover information in child frame."
   :init-value nil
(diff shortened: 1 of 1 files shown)


radon-at-beeper:
@jcs090218 Thoughts on this?

jcs090218 (approved):

jcs090218:
Sorry for the late reply. 😓 

LGTM, merged now! Thank you!

